PR #569, opened 2026-08-02
fix(deepl): send preserve_formatting/split_sentences as JSON booleans

Epoche-Napoleon:
Fixes #568.

## What

`DeepLTranslator` cast `preserve_formatting`/`split_sentences` to the strings `"1"`/`"0"` before building the JSON request body. DeepL's `/v2/translate` accepts that string form only on the classic form-encoded API; over JSON it rejects the request with `{"message":"Value for 'preserve_formatting' not supported."}` — verified against the live API (see #568 for the curl comparison).

Both payload builders now pass the native `bool` through unchanged.

## Why this slipped through

`TranslationOptions` defaults `preserveFormatting` to `true`, so this affected the specialized-translator path on effectively every call that reaches DeepL with default options — not just callers opting into formatting preservation. The existing coverage in `DeepLTranslatorOptionsTest` uses a mock client whose `withBody()` stub discards its argument, so no test could see the actual JSON body being sent.

## Testing

Added `translateSendsPreserveFormattingAndSplitSentencesAsJsonBooleans`, which builds its own request/stream factory stubs (scoped to this one test, not touching the shared `AbstractUnitTestCase` helpers) to capture and `json_decode()` the real request body, then asserts the decoded values are strictly `true`/`false` rather than `"1"`/`"0"`.

Verified locally:
- New test fails against the pre-fix code (reverted the fix, re-ran — confirmed `Failed asserting that '1' is identical to true.`), passes after.
- Full `Tests/Unit/Specialized/Translation/` suite: 242 tests, 0 failures.
- `./Build/Scripts/runTests.sh -s phpstan`: no errors (level 10).
- `./Build/Scripts/runTests.sh -s cgl`: clean (one auto-fixable style nit in the new test, applied).

## Note on commit signing

I don't have a GPG key configured in this environment, so the commit is DCO-signed off (`--signoff`) but not GPG-signed (`-S`). Happy to re-sign or let a maintainer cherry-pick if that's a hard requirement for merge.


CybotTM (changes requested):
Thanks for the precise analysis and the fail-before/pass-after verification — the `preserve_formatting` half of this fix is confirmed: DeepL's own [OpenAPI spec](https://github.com/DeepLcom/openapi/blob/main/openapi.yaml) types it as `PreserveFormattingOption` (`type: boolean`) for the JSON request body, with the `'0'/'1'` string enum (`PreserveFormattingOptionStr`) reserved for the form-encoded variant. Since `sendDeeplRequest()` sends `Content-Type: application/json`, the change is right.

**Blocking: `split_sentences` is not a boolean in the spec — in either body variant.** `SplitSentencesOption` is a string enum `'0' | '1' | 'nonewlines'` for both the JSON and the form-encoded request body. Sending a JSON `true`/`false` contradicts the spec and risks exactly the rejection this PR fixes for `preserve_formatting`. The error message quoted in #568 only covers `preserve_formatting`; unless you have a live-API verification showing DeepL accepts a boolean `split_sentences` over JSON, please keep it as the string `'1'`/`'0'` and adjust the new test's `assertFalse($decoded['split_sentences'])` accordingly. (Side note, out of scope here: `DeepLOptions::splitSentences` being `?bool` means the documented `'nonewlines'` value can't be expressed at all — worth a follow-up issue.)

**SonarCloud quality gate is red** ([7.6% new duplicated lines, threshold 3%](https://sonarcloud.io/dashboard?id=netresearch_t3x-nr-llm&pullRequest=569)) — the inline request/stream factory stubs duplicate existing helper structure. Rather than building them per-test, consider extending the shared `createRequestMock()` helper with an optional body-capture callback: that fixes the very weakness this PR documents (the `withBody()` no-op) for the whole test class, and clears the gate at the same time.

**Copilot's inline comment is valid**: `buildBatchPayload()` is also changed but the new regression test only covers the single-translate path — please add a batch-path assertion so the same regression can't come back through `translateBatch()`.

Two housekeeping notes: the CI workflows on this PR still await first-contributor approval, so the local test/phpstan/cgl results are unverified until a maintainer approves the runs. And re commit signing: the DCO check only enforces the `Signed-off-by` trailer (present), and this repo does not require verified signatures — no blocker there.

CybotTM (commented):
Re-review of 85e4e975. All substantive points from the first round are addressed, and addressed well:

- `split_sentences` is back to the `'1'`/`'0'` string cast with the correct spec rationale (`SplitSentencesOption` has no boolean variant in either body form) — verified against the OpenAPI schema again, and your combined end-to-end browser test on netresearch/t3x-cowriter#134 confirms the whole chain against the live API. The updated CHANGELOG entry now describes exactly what ships.
- The shared `createRequestMock()` gained the by-reference body capture instead of per-test stub forests — that fixes the `withBody()` no-op weakness for every future test in one place, which is the right cut.
- The batch-path regression test (`translateBatchSendsPreserveFormattingAsJsonBoolean`) covers Copilot's point; I've resolved that thread.

One thing still holds the gate: [SonarCloud is at 5.4% new duplicated lines](https://sonarcloud.io/dashboard?id=netresearch_t3x-nr-llm&pullRequest=569) (threshold 3%). The flagged block is `DeepLTranslator.php` 555–581 ↔ 640–662 — the shared head of `buildTranslatePayload()` and `buildBatchPayload()` (formality/glossary/preserve_formatting application). The structure predates this PR, but touching those lines pulled them into the new-code scope, so the gate will not clear on its own. Suggest extracting the common option-application into one private helper (e.g. `applyCommonTranslateOptions(array $payload, DeepLOptions $o): array`) used by both builders — that is also the genuinely better shape, since the next option added would otherwise have to be added twice (this PR's bug existed twice for exactly that reason).

With that extraction this is ready from my side. The CI workflows still await maintainer approval for fork runs — that's on us, not on you.

CybotTM (approved):
Re-review of dcc19e41. The `applyCommonTranslateOptions()` extraction from the previous round is in, which was the last open point — the SonarCloud duplication gate that held this was caused by the two payload builders sharing an untouched head, and that head is now one helper.

I pushed one commit to the branch (`dcc19e41`): `RecastingRemovalRector` flagged the two `json_decode((string)$capturedBody, …)` casts in the new regression tests as redundant — the `assertNotNull()` immediately above each already narrows `?string` to `string`. Verified Rector, PHPStan and the unit suite all pass at PHP 8.2 without the casts, so Rector and PHPStan agree here.

The `split_sentences` reasoning is right and the code matches it: `SplitSentencesOption` is `'0'|'1'|'nonewlines'` in both body forms, so it stays a string cast while `preserve_formatting` becomes a real boolean.

One thing for you, non-blocking: the PR title still says "preserve_formatting/split_sentences as JSON booleans", but only `preserve_formatting` becomes a boolean. Worth correcting so the squashed subject matches what ships.

PR #571, opened 2026-08-02
fix(translation): correct translator priority, allow explicit selection

Epoche-Napoleon:
Fixes #570.

## What

`LlmTranslator::getPriority()` returned `100`, higher than `DeepLTranslator`'s `90`. Per the "higher priority sorts first" contract on `TranslatorInterface::getPriority()`, the universal LLM fallback sorted ahead of every specialized translator — and since `LlmTranslator::supportsLanguagePair()` always returns `true`, `TranslatorRegistry::findBestTranslator()` always returned it first, regardless of DeepL's availability. See #570 for the full trace.

Lowered to `-1000`.

Also adds `TranslationOptions::withTranslator()`: a caller that already resolved a translator via `findBestTranslator()` had no way to act on that choice through `translateWithTranslator()`/`translateBatchWithTranslator()` without pinning an entire `LlmConfiguration` just to reach the explicit-translator branch of `TranslationService::resolveTranslator()`. Mirrors the existing `withConfiguration()` — same clone-and-set pattern, included in `toArray()` under `'translator'`.

## Testing

`TranslatorRegistryTest`'s `findBestTranslator*` tests use hand-ordered `ArrayIterator`s, so they can't catch a real priority regression between the two concrete classes. Added a direct assertion in `LlmTranslatorTest` instead:

```php
self::assertLessThan(DeepLTranslator::getPriority(), LlmTranslator::getPriority());
```

Verified locally:
- Reverted the priority value to `100`, re-ran — new test fails (`Failed asserting that 100 is less than 90.`). Restored the fix — passes.
- Full unit suite: 5751 tests, 0 failures.
- `./Build/Scripts/runTests.sh -s phpstan`: no errors (level 10).
- `./Build/Scripts/runTests.sh -s cgl`: clean, nothing to fix.

## Note on commit signing

No GPG key configured in this environment — commit is DCO-signed off (`--signoff`) but not GPG-signed (`-S`). Happy to re-sign or let a maintainer cherry-pick if required for merge.

## Related

This nr-llm fix is a prerequisite for a follow-up PR I'll open against `netresearch/t3x-cowriter`, which wires its quick-translate action to prefer a specialized translator when one is available (currently it always uses the LLM chat path).


CybotTM (commented):
Both halves of this PR check out against the existing code. The priority inversion is real: the registry sorts via `#[TaggedIterator(defaultPriorityMethod: 'getPriority')]` (higher first), `LlmTranslator::supportsLanguagePair()` always returns `true`, and `findBestTranslator()` takes the first match — so the fallback always won. `-1000` fixes that, and `getAvailable()` preserves the iteration order. And `withTranslator()` needs no service change because `TranslationService::resolveTranslator()` already reads `$options['translator']` as priority 1 and `toArray()` now emits the key — verified, the wiring is complete.

Three suggestions before merge:

- **Copilot's BC comment is right**: the new `$translator` constructor parameter lands before `$beUserUid`/`$plannedCost`, which breaks any external caller using positional arguments for those. Moving it to the end of the parameter list costs nothing.
- **Please add an identifier constant** (e.g. `LlmTranslator::IDENTIFIER = 'llm'`) while you're here — the follow-up PR netresearch/t3x-cowriter#134 currently has to hardcode the `'llm'` string to detect the fallback.
- **Consider also accepting `'auto'` as a source language in `DeepLTranslator::supportsLanguagePair()`** (or a `null` source in `findBestTranslator()`). netresearch/t3x-cowriter#134 calls `findBestTranslator('auto', $target)`, but `'auto'` is not in `SUPPORTED_SOURCE_LANGUAGES` and `normalizeLanguageCode()` doesn't map it — so DeepL can never be selected for auto-detected sources and the cowriter feature is inert without this. DeepL auto-detects when `source_lang` is omitted (which `buildTranslatePayload()` already does for `null`), so treating `'auto'` as supported is faithful to the API. Fine as a follow-up PR if you prefer to keep this one minimal, but it should land before the cowriter PR does.

Copilot's other two comments (hardcoded "90" in the comment/docblock) are cosmetic; the new test referencing `DeepLTranslator::getPriority()` directly — rather than the literal — already keeps the actual invariant honest.

Housekeeping: CI workflows await first-contributor approval, so the stated test/phpstan results are unverified until the runs are approved. Re commit signing: DCO only checks the `Signed-off-by` trailer (present); the repo doesn't require verified signatures, so no blocker. This PR should merge and be released before netresearch/t3x-cowriter#134.

CybotTM (dismissed):
Re-review of 0c681192 — approving. All three requests from the first round are addressed, each in the right place:

- `$translator` moved to the end of the constructor, with a comment that explains the BC constraint to the next person who wants to "group" the fields again. Positional callers of `$beUserUid`/`$plannedCost` are safe.
- `LlmTranslator::IDENTIFIER` added and used internally — netresearch/t3x-cowriter#134 already consumes it.
- The `'auto'`/`''` source handling landed in `DeepLTranslator::supportsLanguagePair()` with the contract documented on `TranslatorInterface` for future implementers — that's more than I asked for and exactly right: DeepL auto-detects when `source_lang` is omitted, so this is faithful to the API, and your end-to-end test (`"translator": "Deepl"` in the real AJAX response) proves the selection chain works in practice now.

SonarCloud is green, the priority regression guard pins the invariant via `DeepLTranslator::getPriority()` rather than a literal, and I've resolved the remaining Copilot threads (the two "hardcoded 90" comments are illustrative prose, the invariant itself is asserted symbolically — no change needed; the constructor-position one is fixed).

The only outstanding gate is CI approval for fork workflows (our side). After merge we'll tag an nr-llm release so #134 can bump its constraint — that release also unblocks the TER/Packagist consumers of `withTranslator()`.

PR #643, opened 2026-08-08
fix(context): count the skill block against the conversation budget

CybotTM:
## Description

`ConversationService::send()` computed the context budget and then dispatched through `LlmServiceManager::chatForConfiguration()`, which injects up to 24 kB of skill block **after** the fit. The budget bound a message list that was never sent that way.

Not a corner case: a criteria-mode configuration has `model_uid = 0`, so `getLlmModel()` is null and `fit()` always falls back to `UNKNOWN_WINDOW_FALLBACK = 8192`. A full block is ~7 900 estimated tokens against a budget of `8192 − 1000 reserve − 246 safety = 6 946`.

`ContextWindowManagerInterface::fit()` gains an optional trailing `string $injectedText` — payload that reaches the wire without being in the message list, the same shape `$toolSpecs` already has. `ConversationService` composes the block once and passes it.

### Deviation from the issue text — please read

The issue says "passes it as a **reserve**", i.e. deduct it from the budget. **That was wrong and is not what this does.**

For exactly the case the issue describes, a deduction pushes the budget non-positive (`6946 − 7896 < 0`), which trips the guard at `Classes/Service/Context/ContextWindowManager.php:78` and returns `passthrough()` — no pruning at all. The fix would have disabled pruning for precisely the class of run it exists for.

The block is therefore charged to the **estimate**, which is what the file already does for the system prompt `MessageShaper` prepends after `fit()`. Drop decisions, `overflowAtFloor` and the recalibration input are identical to a deduction except in that degenerate case; `budget` keeps meaning "window headroom" and `estimatedTokens` keeps meaning "what goes on the wire".

### What was deliberately not done

The injection was **not** moved before the fit. `SkillInjectionService::augmentMessages()` prepends to the *first user message*, and `ContextWindowManager::partition()` counts everything up to and including the first user turn as the never-droppable HEAD. Injecting before the fit would drop the entire history making room and still overflow.

### Known limits, named rather than papered over

- The null-configuration branch of `dispatch()` is not covered — there the manager resolves the installation default itself, and this service never learns which configuration that is.
- The block is composed twice per turn: once here for the estimate, once in the manager for the injection. Both go through the same shared `SkillComposer` DI service (`Configuration/Services.yaml:339`) with identical arguments (`composeBlock($configSkills, [])`), so they agree. Removing the second composition would mean restructuring the injection point, which is a larger change than this fix.

## Related Issue

Closes #625

## Type of Change

- [ ] Documentation update

## Checklist

- [x] I have run `composer ci` and all checks pass — unit, PHPStan (level 10), cgl, fuzzy and the new functional class re-run independently of the authoring pass: all SUCCESS. Rector was verified green under an 8.2 resolve, then the worktree was restored to 8.4; the two resolves cannot coexist locally and CI pins the Rector job to 8.2.
- [x] I have updated the documentation accordingly — docblocks; no user-facing doc change
- [ ] I have added an ADR under `Documentation/Adr/` if this changes the public surface — not needed: `ContextWindowManagerInterface` carries no `@api` marker and is not in `Tests/Unit/Api/api-surface.txt`; `ConversationService` is in the snapshot but constructors are excluded from it by design

### Tests

- unit (`ContextWindowManagerTest`): the block is charged exactly — budget untouched, estimate grows by exactly the block's estimate, message list untouched; an empty block is byte-identical to before (regression); a transcript that fits alone gets pruned once the block counts, and still fits after dropping; the block never enters the undroppable HEAD (guard against the rejected solution).
- unit (`ConversationServiceTest`): the composed block reaches `fit()`; without a composer, `''` is passed.
- functional (`ConversationSkillBudgetTest`, new): real DB session store, real MM-loaded skill, real composer, real `ContextWindowManager`, criteria-mode configuration. 8 seeded turns and a ~12 kB skill body → turns are dropped and the send stays under budget; a control configuration without skills keeps every turn; the stored history is never shortened.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- The conversation context budget counts the skill block (#625).
+  `ConversationService` fitted the transcript and then dispatched into
+  `LlmServiceManager::chatForConfiguration()`, which prepends up to 24 000
+  bytes of skill block to the first user message — the fitted list and the
+  sent list were different lists. A criteria-mode configuration has no model
+  relation, so its window falls back to 8192 tokens, of which a full block is
+  roughly 7 900 estimated tokens against a budget of 6 946 (8192 minus the
+  1000-token response reserve minus the 3 % safety margin): the whole
+  configuration class overran. `ContextWindowManagerInterface::fit()` takes an
+  optional trailing `$injectedText` for payload that reaches the wire outside
+  the message list, and the conversation path composes the block once and
+  passes it. The injection stays where it is — the first user turn is the
+  never-droppable head, so injecting before the fit would drop the entire
+  history and still overflow. Known limit: a session opened without a
+  configuration resolves the installation default inside the manager, so its
+  block is still unaccounted for.
+
 ## [0.26.0] - 2026-08-06
 
 ### Added
--- a/Classes/Service/Context/ContextWindowManager.php
+++ b/Classes/Service/Context/ContextWindowManager.php
@@ -56,6 +56,7 @@ final class ContextWindowManager implements ContextWindowManagerInterface
         ?ChatOptions $options,
         ?UsageStatistics $lastUsage,
         array $toolSpecs = [],
+        string $injectedText = '',
     ): ContextFitResult {
         // A null $lastUsage marks the first send of a run's loop. Reset the
         // per-run calibration state here so a single shared manager instance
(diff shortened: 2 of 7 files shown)


CybotTM (commented):

### Ein Fund, und er betrifft die Merge-Reihenfolge

**Die Commits dieses PRs sind vollständig in #647 enthalten** (`git merge-base --is-ancestor` bestätigt es). #647 steht aber auf Basis `main`, nicht auf diesem Branch — GitHubs Stack-Feature führt ihn als Stack-Wurzel und verweigert das Umhängen.

Folge: **#643 muss vor #647 gemergt werden.** Andernfalls bringt #647 den Code hier mit, und dieser PR bleibt mit leerem Diff zurück. Auf GitHub ist diese Abhängigkeit nirgends sichtbar — #647s Diff zeigt 677 Zeilen, von denen 514 aus #643 stammen, und ein Reviewer sieht denselben Code zweimal.

### Geprüft, ohne Fund

- **Aufschlagen statt Abziehen.** `$outsideTokens` wird auf die Schätzung addiert, nicht vom Budget abgezogen. Der Kommentar nennt den Grund, und er trägt: ein Abzug, der das Budget auf ≤ 0 drückt, löst den Passthrough aus und schaltet das Pruning ganz ab — genau den Überlauf, den die Buchführung verhindern soll.
- **`injectedTokens()` schätzt als User-Nachricht**, rechnet also den Nachrichten-Overhead obendrauf, obwohl der Block *in* eine bestehende Nachricht wandert. Das irrt nach oben — dieselbe Richtung wie der Schätzer sonst auch.
- **Leerer String → 0**, kein Overhead für einen nicht vorhandenen Block.
- Der dokumentierte Restfall (Session ohne Konfiguration löst den Installationsstandard erst im Manager auf, sein Block bleibt ungezählt) steht im CHANGELOG statt zu fehlen.

PR #644, opened 2026-08-08
fix(agent): fail closed on an unrecorded write approval and bind the reviewed turn

CybotTM:
## Description

Two defects on the approval path, fixed together because both sit in `ResumeCoordinator::approve()`.

**The approval decision was not audited.** `AgentRunPersister::recordApproval()` caught `Throwable`, logged a warning and returned `void`; `ResumeCoordinator` called it unchecked and continued into `executeResume()`. A write could be approved and executed with the authorising decision recorded nowhere. The write *step* audit is fail-closed; the decision that authorised it was not.

**The reviewed turn was not bound on every surface.** The stale-review digest existed only in `AgentRunController::approveAction`. `ToolPlaygroundController::resumeAction()` is a registered AJAX route (`nrllm_tool_resume`) and read only `runUuid` and `approve` — nothing bound the decision to the turn under review.

### The subtle part: which state gets resumed

The decode now happens **after** the claim, from the freshly loaded row.

Losing the claim race does not leave the run untouched. The winner runs the turn, the loop continues, and the run can suspend **again on a different turn** — which is exactly the row our claim then succeeds on. The pre-claim read would be the previous turn, so the digest check, the write classification and the execution all have to use the fresh state or they judge a turn that is no longer pending.

### Two gates, both releasing rather than settling

1. The decision must name the turn it was made on. A null digest and a mismatching digest both mean "the reviewed turn is not known", so both are refused (`hash_equals`, not `!==`). This is now the **only** place the invariant lives — the controller comparison is gone.
2. An approval whose APPROVAL event could not be stored may not execute a turn that declares a write. Read-only turns stay fail-soft.

Both release the run back to `WAITING_FOR_APPROVAL` through the existing `suspend()` transition, which clears claim and lease. Nothing ran; the operator can decide again. A release that itself fails settles the run rather than stranding it RUNNING, invisible to both the inbox and the reaper.

**A denial passes gate 1 but deliberately not gate 2.** Gate 2 stops an unaudited *write* from executing, and a denial executes nothing. Refusing it would leave the write-declaring turn pending and approvable while the operator who wanted it gone is turned away. The lost "who denied" record is logged. A test pins the asymmetry.

### Corrections to the issue text — please read

- **`Tests/Unit/Api/api-surface.txt` does change.** The issue said constructors are exempt from the snapshot, so none was expected. Constructors are — *promoted properties* are not. Exactly one line was added: `property readonly turnDigest: ?string`. Regenerated the documented way.
- **The playground has no approval UI.** The issue asked for the playground JS to send the digest back. `Resources/Public/JavaScript/` contains no `approve`, `runUuid` or `resume` — the handler exists, the client does not. No JS was changed. The server now emits the digest in both pause shapes; binding the reachable endpoint is the point, and building an approval UI is a different task.
- `ResumeCoordinator` has no explicit `Services.yaml` entry; namespace autoregistration covers it, so the new collaborators autowire without a config change.

## Related Issue

Closes #621

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

`ApprovalDecision` is `@api` and constructed by third parties. The third parameter is optional in the signature but **mandatory in behaviour**: a null digest is refused. That is deliberate — silently accepting an unbound decision would defeat the gate — and it is called out in CHANGELOG.

## Checklist

- [x] I have run `composer ci` and all checks pass — unit, PHPStan (level 10), cgl, fuzzy and both functional controller classes re-run independently of the authoring pass: all SUCCESS (functional: 22 tests, 364 assertions). Rector was verified green under an 8.2 resolve, then the worktree was restored to 8.4.
- [x] I have added an ADR under `Documentation/Adr/` if this changes the public surface — ADR-132

### Tests

- unit: `PendingTurnDigest` is byte-identical to the previous inline computation, including the `serialize()` fallback (characterisation).
- unit: `recordApproval()` returns false when the event store throws.
- functional: a stale digest is refused on **both** surfaces and the run is `WAITING_FOR_APPROVAL` afterwards — not RUNNING, not terminal. A null digest is refused likewise.
- functional: a write turn whose approval event cannot be stored executes nothing and is decidable again (new `ApprovalEventFailingRunRepository` fixture — a real DB repository where only that one event kind fails).
- functional: a read-only turn continues despite a failed approval event; a **denied** write turn also continues, pinning the asymmetry.
- `AgentRunControllerTest` now drives `approveAction` including PRG and flash mapping, rather than deferring to unit tests.

### Note for reviewers

`Documentation/Adr/Index.rst` will show gaps around ADR-133..137 until the sibling branches land.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,33 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- **The approval of a write is fail-closed and bound to the turn that was
+  reviewed** (ADR-132). Two defects on the same path:
+  `AgentRunPersister::recordApproval()` swallowed a store error and
+  `ResumeCoordinator::approve()` executed anyway, so a write could run with no
+  record of who authorised it; and the stale-review digest existed only in the
+  `AgentRunController`, so the Tool Playground could approve a turn it had never
+  displayed. `recordApproval()` now returns `bool` (same shape as
+  `recordStep()`), and `approve()` refuses to execute a turn declaring a write
+  whose decision could not be recorded. The verified digest moved into
+  `ApprovalDecision` and is compared — timing-safe — against the state loaded
+  AFTER the resume claim, so a decision made on a turn a concurrent approval
+  already replaced is refused as well. Read-only turns stay fail-soft
+  (deliberately), and a denial still passes: it executes nothing, so refusing it
+  would only leave the write pending. A refused decision RELEASES the run back
+  to `WAITING_FOR_APPROVAL` — nothing runs, nothing settles, the operator can
+  re-review and decide again. The playground's `awaiting_approval` payload and
+  streaming event now carry `turnDigest`, and `resumeAction()` reads it back.
+  **Breaking:** `ApprovalDecision`'s constructor takes a third argument,
+  `?string $turnDigest`. It is optional in the signature for source
+  compatibility but MANDATORY at runtime — a `null` is refused exactly like a
+  wrong digest, so any third-party caller that constructs the decision itself
+  must supply the digest of the turn it displayed or every approval will fail
+  with `StaleApprovalTurnException`. `WaitingRunViewFactory::pendingTurnDigest()`
+  is gone; the computation lives in the new `PendingTurnDigest` service.
+
 ## [0.26.0] - 2026-08-06
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   wrong digest, so any third-party caller that constructs the decision itself
   must supply the digest of the turn it displayed or every approval will fail
   with `StaleApprovalTurnException`. `WaitingRunViewFactory::pendingTurnDigest()`
-  is gone; the computation lives in the new `PendingTurnDigest` service.
+  and `WaitingRunViewFactory::turnDigestForRun()` are gone; the computation lives
+  in the new `PendingTurnDigest` service, and the rendered card is the only
+  carrier of the value. (Neither was tracked public API.)
+  A run whose `suspended_state` is already unreadable is still refused BEFORE
+  the resume claim, so it stays `WAITING_FOR_APPROVAL` with its state intact
+  instead of being claimed and settled `FAILED` — the guarded terminal settle
+  clears `suspended_state`, which would destroy the only copy a repair could
+  work from. The decode after the claim remains and now covers only the race it
+  is there for: the state CHANGED between the two reads and the row the claim
+  won is the unreadable one.
 
 ## [0.26.0] - 2026-08-06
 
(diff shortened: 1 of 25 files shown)


CybotTM (commented):

Copilot antwortet auf diesem PR mit `Copilot was unable to review this pull request because the user who requested the review has reached their quota limit.` Das ist keine Bewertung, sondern ein Kontingentfehler. Ich habe den Diff daher selbst gelesen und sage das hier ausdrücklich, damit niemand die COMMENTED-Zeile für eine Review hält.

**Einschränkung vorweg:** Ich bin der Autor dieses PRs. Ein Selbst-Review findet systematisch weniger als ein fremdes — es ersetzt die ausstehende unabhängige Durchsicht nicht.

### Geprüft

- Die Reihenfolge der beiden Decodes und was jeder refüsiert (nicht-destruktiv vor dem Claim, settle danach).
- Beide Fail-Closed-Gates und der Release-Pfad, inklusive des Falls, dass der Release selbst scheitert.
- `turnDeclaresWrite()`: unbekanntes Tool → `NON_IDEMPOTENT_WRITE`, unlesbarer Eintrag → `true`, Null-Resolver → `true`. Alle drei Richtungen fail-closed.
- Die hinzugefügte `$claimed instanceof AgentRun`-Prüfung schließt eine echte Null-Dereferenzierung.

### Ein Fund

`ResumeCoordinator::approve()` begründet `hash_equals` mit:

> the digest is attacker-supplied and compared against a **server secret-ish value**

Das stimmt nicht. `PendingTurnDigest::forState()` ist ein schlichter sha256 über die Pending-Calls, und die Karte rendert ihn dem Betrachter. Der Digest ist **nicht geheim** — wer den Turn sieht, kann ihn ausrechnen. Er schützt gegen *Veraltung* (stale tab, zweiter Operator), nicht gegen Fälschung.

`hash_equals` bleibt richtig und kostet nichts. Falsch ist die Begründung, und eine Kommentarzeile, die eine Sicherheitseigenschaft behauptet, die es nicht gibt, ist genau die Sorte Aussage, auf die sich der nächste Leser stützt. Ich korrigiere den Kommentar.

### Kein Fund, aber notiert

`turnDeclaresWrite()` liefert `false` für eine leere `pendingCalls`-Liste, Gate 2 entfällt dann. Das ist konsistent: eine leere Liste führt nichts aus, es gibt also nichts zu verhindern — dieselbe Begründung, mit der eine Ablehnung Gate 2 passiert.

CybotTM (commented):
Re-Review nach dem Nachziehen auf main (Runde 2).

Der Push war ein Merge von `main` plus die Auflösung zweier additiver Konflikte — CHANGELOG (beide Einträge behalten, eine `### Fixed`-Überschrift) und `Documentation/Adr/Index.rst` (Vereinigung, nach ADR-Nummer sortiert). Kein Produktionscode geändert.

Gates auf dem neuen Head: PHPStan, Unit und CGL grün. Copilot ist weiterhin am Kontingentlimit, die COMMENTED-Zeile ist ein Fehler und keine Bewertung.

PR #646, opened 2026-08-08
test(context): pin the assembly order on all three message-list paths

CybotTM:
## Description

Assembly order is currently an emergent property of several hard-wired paths, and the invariant that keeps a configuration's system prompt alive is stated only in a code comment. A comment is not a test, and #638 changes one of these paths.

**Characterisation only — no production code is touched.** A defect found while reading is reported, not fixed (see below).

### Paths covered

- **Agent loop, four variants**: no augmentation (the production case, behind the early return at `ToolLoopService.php:659-661`), playground, dry run, codec resume.
- **Conversation**: `system → history oldest-first → new user turn`, one system message however many turns.
- **Streaming**: `system(config prompt) → user(skill block + turn)`. The two injection sites sit ~600 lines apart in the manager; the resulting order is now stated rather than inferred.

### The load-bearing invariant, asserted in both directions

The system prompt must precede all system messages, or `MessageShaper::applySystemPrompt()`'s guard fires and the configuration's prompt is silently dropped for that run.

- Forward: with the baked prompt in the lead, the shaper returns the list unchanged and position 0 is the configuration prompt.
- Counterfactual (`aSnippetSystemMessageAheadOfTheSystemPromptWouldDropItSilently`): a snippet ahead of the prompt satisfies the guard, and the configuration prompt is absent from every message.

Also pinned: the dry run assembles byte-for-byte what a live run sends (`assertSame` between the trace snapshot and the live provider argument) and calls no provider; `skipAssembly` leaves the transcript byte-identical; the skill block never enters the system role.

### The tests were verified to fail meaningfully

Moving the snippet loop ahead of the system-prompt bake in `assemble()` turned three tests red on identity assertions. Probe reverted; the file is byte-identical to its backup.

## Related Issue

Closes #637

Found while writing these and filed separately, not fixed here: #645 — `SkillInjectionService::isUserMessage()` does not recognise a multimodal user message, so the skill block lands on a later turn or is silently dropped.

## Type of Change

- [ ] Documentation update

Tests only — none of the above categories fit.

## Checklist

- [x] I have run `composer ci` and all checks pass — unit, PHPStan (level 10) and cgl re-run independently of the authoring pass: all SUCCESS. The three new classes: 15 tests, 62 assertions. Also run under `-p 8.2` (PHPUnit 11.5) with no deprecations, so they are cross-version safe.
- [ ] I have updated the documentation accordingly — not applicable
- [ ] I have added a `CHANGELOG.md` entry under `## [Unreleased]` — not applicable, no behaviour change
- [ ] I have added an ADR under `Documentation/Adr/` if this changes the public surface — not applicable

### Correction to the issue text

The issue put the conversation assembly at `ConversationService.php:100-153`. The assembly is `100-114`; `116-153` is the ADR-121 context-window fit, which replaces the assembled list wholesale and is not part of the ordering question.

### DIFF
--- a/Tests/Unit/Service/Feature/ConversationServiceAssemblyOrderTest.php
+++ b/Tests/Unit/Service/Feature/ConversationServiceAssemblyOrderTest.php
@@ -1,0 +1,222 @@
+<?php
+
+/*
+ * Copyright (c) 2025-2026 Netresearch DTT GmbH
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+declare(strict_types=1);
+
+namespace Netresearch\NrLlm\Tests\Unit\Service\Feature;
+
+use Netresearch\NrLlm\Domain\Model\CompletionResponse;
+use Netresearch\NrLlm\Domain\Model\LlmConfiguration;
+use Netresearch\NrLlm\Domain\Model\UsageStatistics;
+use Netresearch\NrLlm\Domain\Repository\LlmConfigurationRepository;
+use Netresearch\NrLlm\Domain\ValueObject\AiActorContext;
+use Netresearch\NrLlm\Domain\ValueObject\ChatMessage;
+use Netresearch\NrLlm\Service\ConfigurationResolver;
+use Netresearch\NrLlm\Service\Feature\ConversationService;
+use Netresearch\NrLlm\Service\LlmServiceManagerInterface;
+use Netresearch\NrLlm\Service\MessageShaper;
+use Netresearch\NrLlm\Service\Option\ChatOptions;
+use Netresearch\NrLlm\Tests\Unit\Service\Session\Fixtures\RecordingAiSessionRepository;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+
+/**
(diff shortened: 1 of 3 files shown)


CybotTM (commented):

Reiner Test-PR: 939 Zeilen, drei Klassen, kein Produktionscode.

### Worauf ich bei einem Charakterisierungs-PR schaue

Ob die Tests das prüfen, was sie zu prüfen behaupten — ein Test, der nur Mock-Verhalten zurückliest, nagelt nichts fest.

`StreamAssemblyOrderTest` fängt die Nachrichten am Provider ab und behauptet dann über die **echte** Zusammensetzung: `assertSame(['system','user'], $this->roles())`, der Systemprompt als Inhalt der ersten Nachricht, der Skill-Block *innerhalb* der User-Nachricht. Gemockt sind die Ränder (Provider, Repository, Cache, Logger) — zusammengesetzt wird von `LlmServiceManager` und `SkillInjectionService` selbst. Das ist die richtige Aufteilung.

Der Reihenfolge-Aspekt ist auch der Grund, warum diese Tests wertvoll sind: Sie halten fest, in welcher Reihenfolge Systemprompt, Skill-Block und Verlauf entstehen — eine Eigenschaft, die kein Typ erzwingt und die bei jeder Umstellung der Naht stillschweigend kippen kann.

Kein Fund.

PR #647, opened 2026-08-08
docs(adr): context assembly is a seam, not a provider registry

CybotTM:
## Description

ADR-139. Records why nr_llm does **not** get a `ContextProviderInterface` with a registry, and names what would change that.

A programme proposal asked for `supports()`/`provide()`, a DI-tagged registry, a `ContextRequest` (actor, site, language, page, requested types), a `ContextResult` (items, classification, source, freshness, token estimate) and two core reference providers. The motivation is sound; the shape is not. Four things about the code decided it:

- **No vocabulary problem to solve.** Skills carry instructions, snippets carry fragments, and ADR-031 deliberately keeps snippet tags a *convention* between editors and consumers rather than a core enum. A core-side vocabulary of context types would move editorial semantics into the layer whose point is not to have them.
- **No entry point could fill the request.** Nothing carries site, language and page. The proposed reference providers would have to find the page themselves — a generic read-everything accessor, which is a standing non-goal.
- **`freshness`, `validUntil`, `source` have no reader.** Nothing caches, invalidates or displays provenance.
- **The budget problem is real but not registry-shaped.** The skill block is bounded in bytes, the transcript in tokens, and the second did not know about the first. That was a defect and was fixed as one (#643). It needed the two budgets to meet in the one calibrated estimator, not a budget object with `max provider count`.

### Deferred with named triggers, not rejected

- **Provider registry** — when a consuming extension has a source expressible neither as a skill nor as a tagged snippet. Designed against that source, not against a guess.
- **A data classification for injected context** — skills, snippets, system prompt and task input carry no `ToolDataClass` and are checked against no trust zone. Deferred rather than ignored because that content already passes the input-guardrail screener on the way out (ADR-087) and secret redaction is a **mandatory** guardrail a configuration cannot select away; screening runs on the assembled list *after* skill injection, so the block is screened too. What is missing is the classification, not every check. Building it means an operator-declared column, an ADR-115-shaped migration and a new governance decision case — a fail-closed axis over existing data, not something to add speculatively.

### Written down rather than closed

`LlmServiceManager` binds **no** context window at all — its chat, completion and streaming paths inject skills and send. Only `ConversationService` and `ToolLoopService` call `ContextWindowManager`. A long transcript through the generic API is bounded by the provider, not by us. Named in the ADR so the seam is not mistaken for full coverage.

## Related Issue

Closes #636

## Type of Change

- [x] Documentation update

## Checklist

- [ ] I have run `composer ci` and all checks pass — not applicable, no code touched; CI renders the docs
- [ ] I have added tests that prove my fix/feature works — not applicable
- [ ] I have added a `CHANGELOG.md` entry under `## [Unreleased]` — not applicable, no user-facing change
- [x] I have added an ADR under `Documentation/Adr/` if this changes the public surface — ADR-139, entered in the toctree

### Claims checked against the tree before writing them down

- All six `:ref:` targets resolve (`adr-031`, `adr-035`, `adr-087`, `adr-115`, `adr-120`, `adr-122`).
- ADR-031 does keep the tag vocabulary a convention (`Adr031PromptSnippetLibrary.rst:53-61`), including the exact-token match where `style` never matches `lifestyle`.
- `SecretRedactionInputGuardrail::isMandatory()` exists, and `ProviderContentFilterGuardrail.php:34` records that secret redaction stays mandatory while the content filter can be selected away.
- Skills are injected at `LlmServiceManager.php:817` and `screenInput()` runs downstream of that, so the screening claim holds in that order.
- `grep -c ContextWindowManager Classes/Service/LlmServiceManager.php` → **0**, which is the basis for the "binds no window" statement.

One correction made during review: the deferral originally cited ADR-098, which covers input screening for *specialized* prompts. The chat send path is ADR-087. Fixed before commit.

### Note for reviewers

`Documentation/Adr/Index.rst` will show gaps at ADR-132..138 until the sibling branches land.

### DIFF
--- a/Documentation/Adr/Adr139ContextAssemblyIsASeam.rst
+++ b/Documentation/Adr/Adr139ContextAssemblyIsASeam.rst
@@ -1,0 +1,162 @@
+.. include:: /Includes.rst.txt
+
+.. _adr-139:
+
+============================================================================
+ADR-139: Context assembly is a seam, not a provider registry
+============================================================================
+
+:Status: Accepted
+:Date: 2026-08-09
+:Authors: Netresearch DTT GmbH
+
+.. _adr-139-context:
+
+Context
+=======
+
+A programme proposal asked for a :php:`ContextProviderInterface` with
+``supports()`` / ``provide()``, a DI-tagged registry, a :php:`ContextRequest`
+carrying actor, site, language, page and requested types, a
+:php:`ContextResult` carrying items, classification, source, freshness and a
+token estimate, and two reference providers in the core.
+
+The motivation is sound: consuming extensions want to feed editorial context —
+brand voice, audience, legal instructions — into a request without nr_llm
+knowing what any of that means.
+
+The shape is not. Building it now would produce an interface with zero
+implementers and zero readers, which is the pattern :ref:`adr-120` removed
+once already: an argument that looked like enforcement, was read by nothing,
+and bought false trust.
+
+Four things about the code decided this.
+
+**There is no vocabulary problem to solve.** Two mechanisms already carry
+editorial context. Skills (:ref:`adr-035`) carry instructions. Snippets
+(:ref:`adr-031`) carry fragments and are addressed by free-form tags that
+:ref:`adr-031` deliberately keeps as a *convention* between editors and
+consumers, not a core enum. A core-side vocabulary of context types — ``brand``,
+``audience``, ``editorial`` — would move editorial semantics into a layer whose
+whole point is not to have them.
+
+**There is no entry point that could fill the request.** A
+:php:`ContextRequest` with site, language and page presumes callers that carry
+those. None does. The reference providers the proposal names
+(``CurrentPageContextProvider``, ``CurrentSiteContextProvider``) would have to
+find the page themselves — which is a generic read-everything accessor, a
+standing non-goal.
+
+**``freshness``, ``validUntil`` and ``source`` have no reader.** Nothing caches
+context, nothing invalidates it, nothing displays where it came from.
+
+**The budget problem is real, and it is not shaped like a registry.** The two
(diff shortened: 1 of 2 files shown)


CybotTM (commented):

### Merge-Reihenfolge: #643 zuerst

Dieser PR **enthält die Commits von #643**. Sein Diff gegen `main` zeigt 677 Zeilen, davon sind 514 der Code aus #643; eigen sind nur die 163 Zeilen ADR-139 plus der Index-Eintrag.

Basis auf `fix/skill-block-in-conversation-budget` umhängen geht nicht — GitHub lehnt es ab, weil dieser PR Wurzel eines nativen Stacks (→ #659) ist. Damit bleibt die Abhängigkeit auf GitHub unsichtbar und muss beim Mergen von Hand eingehalten werden: **#643, dann #647, dann #659.** Andersherum bleibt #643 mit leerem Diff zurück.

### Zum ADR selbst

ADR-139 beschreibt genau die Naht, die #643 im Code zieht, und begründet, warum die Zählung beim Sender liegt und nicht beim Zusammensetzer. Die Aussagen decken sich mit dem, was ich in #643 gelesen habe — insbesondere die Begründung fürs Aufschlagen statt Abziehen.

Ein Hinweis für später: #659 hat dieser Naht inzwischen ein **zweites** nachgestelltes Argument hinzugefügt (`$effectiveSystemPrompt`). Die beiden sind nicht dasselbe — eines wird der Liste *hinzugefügt*, das andere *ersetzt* den abgeleiteten Prompt — und stehen jetzt gemeinsam in der Signatur. Ein drittes wäre der Punkt, an dem die Naht ein Wertobjekt braucht statt einer weiteren Stelle.

Kein Fund am ADR.

PR #648, opened 2026-08-08
fix(skills): make the skill-block byte budget configurable

CybotTM:
### The issue's premise was wrong — please read this first

#626 said `SkillComposerFactory::create()` never sets `maxBytes`, so the 24 000-byte cap is declared and ineffective. It is not. On `main`:

```php
// SkillComposerFactory.php:40
return new SkillComposer(minTrustLevel: $this->resolveMinTrustLevel());

// SkillComposer.php:60-63
public function __construct(
    private int $maxBytes = self::DEFAULT_MAX_BYTES,   // 24000
    private SkillTrustLevel $minTrustLevel = SkillTrustLevel::UNTRUSTED,
) {}
```

`minTrustLevel:` is a **named** argument, so `maxBytes` takes its default and `composeBlock()` enforces it. The factory is the only production construction path (`Configuration/Services.yaml:339-340`), so every instance has always been capped.

The wrong claim came out of a programme audit; it is corrected on the issue too, so it does not get repeated.

### What the actual defect is

The cap was **unadjustable**. A new `skills.maxBytes` key makes it settable.

A characterisation test (`defaultCapIsEnforcedWhenNothingIsConfigured`) pins the pre-existing behaviour so the distinction stays visible in the suite rather than only in this description.

### Fallback direction

Missing, unreadable, non-numeric or non-positive → `SkillComposer::DEFAULT_MAX_BYTES`. This is deliberately the **opposite** direction from the trust floor next to it: a bad trust value must not raise the bar, a bad budget must not remove one. `0` means "use the default", never "uncapped" — matching `rerankerTimeout` and `privacy.retentionDays`, and keeping an emptied field from silently putting an unbounded block on the wire. There is no way to switch the cap off; lowering the value is how you tighten it.

### A second inaccuracy, in the ADR

ADR-036 §5 stated that *"with `Model::contextLength == 0` the absolute cap applies"*, which implies a window-derived ceiling with 24 000 as its fallback. **That derivation never existed in any form** — the ADR documented a feature that was never built. Corrected.

### The window-derived ceiling is deliberately not built

Call sites do have an `LlmConfiguration`, so "no call site" is not the reason. Four stronger ones:

1. **It would break #643.** That branch's `ConversationService::skillBlockFor()` calls `composeBlock()` **directly**, bypassing `SkillInjectionService`, and its docblock states the invariant that both compositions yield the same block. A ceiling threaded through the injection service would apply on the manager's path but not on the measuring call — the invariant would silently become false. (Branch fetched and read, not assumed.)
2. **The window would often be the wrong one.** `getLlmModel()` is null in criteria mode, where the model is picked at runtime.
3. **`ContextWindowManager` already does this, better** — a calibrated token estimator over the real send, which after #643 counts the skill block.
4. `SkillComposer` is a shared readonly service with `maxBytes` fixed at construction; a per-configuration ceiling means changing `composeBlock()` and threading it through three `SkillInjectionService` methods.

Recorded as an explicit consequence in ADR-036 rather than left as silence.

## Related Issue

Closes #626

## Type of Change

- [ ] Documentation update

New configuration key; existing installations are unaffected because the default equals the previously hard-coded value.

## Checklist

- [x] I have run `composer ci` and all checks pass — unit, PHPStan (level 10), cgl and fuzzy re-run independently of the authoring pass: all SUCCESS. New factory tests: 17 tests, 45 assertions. Rector green under an 8.2 resolve, then restored to 8.4 and unit + PHPStan re-run on that final state.
- [x] I have updated the documentation accordingly — `Documentation/Administration/Skills.rst` beside `skills.minTrustLevel`; ADR-036 §5 correction
- [ ] I have added an ADR under `Documentation/Adr/` if this changes the public surface — not needed; ADR-036 is corrected instead

### Worth flagging

An existing guard caught a real bug in the first draft: `PrivacyRetentionConfigurationTest::noSettingLabelContainsASemicolon` failed because the new label contained a `;`, which TYPO3 treats as the field separator and which would have silently truncated the operator-visible description. Replaced with an em dash. The guard earned its place.

Also ran the TYPO3 render-guides image over `Documentation/`: 192 files, zero warnings for either touched file, both new `:ref:` targets resolve.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,36 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- New extension configuration key `skills.maxBytes` (default `24000`) for the
+  byte budget of the composed skill block (ADR-036 §5). `SkillComposer` had
+  accepted `maxBytes` as a constructor argument since the feature landed, but
+  `SkillComposerFactory` — the only production construction path — never
+  passed it, so the ceiling was fixed at the hardcoded default with no way for
+  an operator to change it. The cap itself was never absent: the constructor
+  default applied, so an unconfigured instance was capped at 24 000 bytes then
+  and is capped at 24 000 bytes now. What changes is that it can be adjusted.
+  A missing, unreadable, non-numeric or non-positive value falls back to
+  24 000, so an emptied or fat-fingered field cannot remove the cap — `0`
+  means "use the default", not "uncapped". This is the opposite fallback
+  direction from `skills.minTrustLevel`, which fails towards the *lower* bar
+  so a bad value cannot silently hide skills; a bad budget must not silently
+  unbound the block. Lower `skills.maxBytes` to reserve more of the model's
+  context window for the conversation itself.
+
+### Fixed
+
+- ADR-036 §5 claimed a ceiling derived from the model's context window ("with
+  `Model::contextLength == 0` the absolute cap applies"). No such derivation
+  was ever implemented. The section now says what the code does — the budget
+  is instance-wide and window-independent — and records why deriving it per
+  configuration is deliberately not done: `SkillComposer` is one shared
+  service whose budget is fixed at construction, the configuration's
+  `llmModel` is null in criteria selection mode (so the window read would not
+  belong to the model that serves the call), and `ContextWindowManager`
+  (ADR-107) already bounds the real send with a calibrated token estimator.
+
 ## [0.26.0] - 2026-08-06
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,18 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   `llmModel` is null in criteria selection mode (so the window read would not
   belong to the model that serves the call), and `ContextWindowManager`
   (ADR-107) already bounds the real send with a calibrated token estimator.
+- The `allowed-tools` documentation claimed the tool union and the injected
+  prompt block are the same set (`Tools.rst`, ADR-038 §5,
+  `AllowedToolsResolver`'s docblock: "exactly what SkillComposer injects").
+  They can differ: `effectiveSkills()` does not know the byte budget, which is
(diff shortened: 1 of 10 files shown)


CybotTM (commented):

Kontingentlimit; die COMMENTED-Zeile ist ein Fehler, keine Bewertung. Diff selbst gelesen, als Autor.

### Geprüft

- **Fallback-Richtung.** `resolveMaxBytes()` fällt bei fehlend, unlesbar, nicht-numerisch und `<= 1` auf `DEFAULT_MAX_BYTES` zurück. Das ist die *umgekehrte* Richtung zum Trust-Floor, und der Kommentar sagt auch warum: ein kaputter Trust-Wert darf die Hürde nicht anheben, ein kaputtes Budget darf keine entfernen. Besonders `0` heißt „Default", nicht „unbegrenzt" — ein geleertes Feld lässt so keinen unbegrenzten Skill-Block auf die Leitung.
- **Zahlenrand.** `is_numeric` vor `(int)`; negative Werte und `0` fallen auf den Default, `"1e5"` wird unter PHP 8 korrekt zu 100000. Keine Lücke gefunden.
- **`skillsConfig()` wirft weiter durch** und jeder Aufrufer fängt mit seinem eigenen Fallback — die beiden Fallbacks bleiben dadurch getrennt, was angesichts ihrer gegenläufigen Richtung wichtig ist.
- **`DEFAULT_MAX_BYTES` wird public.** `ApiSurfaceSnapshotTest` erfasst öffentliche Konstanten über `getReflectionConstants(IS_PUBLIC)`; die Unit-Suite ist auf diesem Branch grün, der Snapshot also unberührt.

### Die interessanteste Entscheidung steht im `AllowedToolsResolver`

Das Budget wird dort bewusst **nicht** mitgezählt. Der Grund ist nicht Bequemlichkeit: Fiele der letzte deklarierende Skill aus dem Budget, lieferte der Resolver `null` — „keine Einschränkung", also *jedes* registrierte Tool. Ein strengeres Budget würde das Gate damit **weiten** statt zu verengen. Die Asymmetrie ist im Docblock festgehalten, und sie gehört dorthin.

Kein Fund.

PR #653, opened 2026-08-08
feat(routing): match the resolved model against the operation in criteria mode

CybotTM:
## Description

Criteria mode matched capabilities only against **stored** criteria, never against the operation actually running. A configuration whose criteria do not mention tools could resolve a model that cannot do tools, and the failure surfaced as a provider error rather than a configuration one.

A `ProviderOperation → ModelCapability` map is merged into the criteria before `findMatchingModel()`.

### It carries a pre-existing bug fix it cannot work without — #652

**`Model::$capabilities` never reached the entity.** Extbase resolves property types through Symfony PropertyInfo; `ReflectionExtractor` infers a *collection* from an adder/remover pair, and `addCapability()` / `removeCapability()` inflect to `$capabilities`, so it resolved as `array`. `DataMapper::thawProperties()` has no array mapping, so the column was silently dropped on every load.

Every repository-loaded model came back with an **empty** capability set — which means **every `capabilities` selection criterion has been matching nothing in production**, long before this ticket. Unit tests could not catch it: they construct `Model` with setters, the one path that bypasses the DataMapper.

Fixed with an explicit `@var string`. **The tag needs its description** — PHP-CS-Fixer's `no_superfluous_phpdoc_tags` strips a bare `@var string` as redundant, which would silently reintroduce the bug. The docblock says so.

Verified as a controlled before/after on the same tree rather than by inference: docblock removed → 2 failures (`null is not an instance of Model`); docblock restored → 5 tests, 23 assertions, SUCCESS.

Filed as #652 so the defect has its own record; it is older and wider than this feature.

### Three deviations from the issue text

1. **Only `chat`, `vision` and `tools` are enforced** — not "every operation that implies a capability". No model discoverer writes `completion` or `embeddings`; all seven write `chat[,vision,tools,streaming,reasoning]`. Only four of seven write `streaming` (Ollama, Groq and OpenRouter do not, and their models stream fine). Enforcing those would refuse every wizard-created model on the *primary* completion path and break every criteria-mode embedding configuration. That is an outage, not a check. Recorded in ADR-138 with a revisit condition.
2. **The derived capability goes into its own criteria key**, not into `capabilities`. "Empty CSV means undeclared" and strict matching of operator-stated criteria are incompatible in one list: a model declaring nothing must pass the *derived* check while still failing an *explicit* requirement.
3. **`chat()` / `complete()` / `streamChat()` do reach `resolveModel()`** — the issue says the generic path never does. They prefer the default DB configuration and delegate to `*WithConfiguration()`. Only their ad-hoc pinned-provider branch, plus `embed()`, `vision()` and `chatWithTools()`, bypass it. The boundary is real but narrower than stated, and ADR-138 states it precisely. Those paths are untouched.

### As instructed and confirmed unchanged

`UnsupportedFeatureException` still falls to `FailureClass::UNKNOWN` (`FailureClassifier.php:45`) and is therefore not retryable — a misconfiguration must not be masked by a fallback. The reason is in the ADR so nobody "fixes" it later.

`LlmServiceManager.php:634` and `:653` both pass `ProviderOperation::Embedding`, with a unit test asserting the two resolutions receive the same operation and return the same model — the cache-key trap the issue warned about.

## Related Issue

Closes #635 · fixes #652

## Type of Change

- [ ] Documentation update

Internal signatures change (`ModelSelectionServiceInterface::resolveModel()`, `ConfigurationCallPlanner::resolveModel()`/`adapterFor()`); neither is in the frozen `api-surface.txt`, and no new type enters a frozen signature.

Note that #652's fix makes `capabilities` criteria **start working**. An installation with criteria that were silently ignored will now see them applied — the new `routing.operationCapabilityEnforcement` switch (ADR-113/115 shape, only a literal `observe` observes) exists for the derived check; the pre-existing criterion has no such switch because it was always meant to apply.

## Checklist

- [x] I have run `composer ci` and all checks pass — unit, PHPStan (level 10), cgl and fuzzy re-run independently of the authoring pass: all SUCCESS; the functional classes executed for real (45 tests, 485 assertions). Rector green under an 8.2 resolve, then restored to 8.4 and the suites re-run on that tree.
- [x] I have added an ADR under `Documentation/Adr/` if this changes the public surface — ADR-138

### Named, not fixed

`ConfigurationController::testConfigurationAction()` catches `ProviderException` — the parent of `UnsupportedFeatureException` — and answers 502 "LLM provider error". An admin testing a misconfigured criteria configuration now gets a provider-shaped message for a configuration-shaped problem. Fixing it needs new locallang keys and is a separate change; naming it here rather than claiming it is filed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,39 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- Criteria-mode configurations no longer resolve a model that cannot serve
+  the running operation (ADR-138). The operation is merged into the criteria
+  before selection, so a tool call picks a tool-capable model instead of
+  failing later as a provider error. Only `chat`, `vision` and `tools` are
+  enforced — no model discoverer writes `completion` or `embeddings`, and
+  only four of seven write `streaming`, so requiring those would refuse
+  working models. A model whose capability field is empty counts as
+  undeclared and always passes. New extension setting
+  `routing.operationCapabilityEnforcement` (default `enforce`, fail-closed
+  like ADR-113) switches the axis to `observe`, which logs the mismatch and
+  leaves selection untouched. Fixed-mode configurations are unaffected.
+- `Model::$capabilities` was dropped on every load from the database.
+  Extbase resolved the property as an array (Symfony PropertyInfo infers a
+  collection from the `addCapability()` / `removeCapability()` pair) and its
+  DataMapper has no array mapping, so every repository-loaded model came
+  back with an empty capability set — which also made every `capabilities`
+  selection criterion match nothing. Anything reading `getCapabilities()`
+  off a repository-loaded model now sees the persisted value.
+- The embedding cache key and the embedding call can no longer resolve
+  different models for the same call. `embedForConfiguration()` resolves
+  twice — once outside the pipeline for the key, once inside the terminal —
+  and both now pass the same operation.
+
+### Changed
+
+- Internal signature change: `ModelSelectionServiceInterface::resolveModel()`
+  and `ConfigurationCallPlanner::resolveModel()` take a required
+  `?ProviderOperation`; `ConfigurationCallPlanner::adapterFor()` likewise.
+  Neither class is part of the frozen `@api` surface. Callers with no
+  operation pass `null` explicitly.
+
 ## [0.26.0] - 2026-08-06
 
 ### Added
--- a/Classes/Controller/Backend/ConfigurationController.php
+++ b/Classes/Controller/Backend/ConfigurationController.php
@@ -23,6 +23,7 @@ use Netresearch\NrLlm\Domain\Repository\LlmConfigurationRepository;
 use Netresearch\NrLlm\Domain\Repository\ModelRepository;
 use Netresearch\NrLlm\Provider\Exception\ProviderException;
 use Netresearch\NrLlm\Provider\Exception\ProviderResponseException;
+use Netresearch\NrLlm\Provider\Middleware\ProviderOperation;
 use Netresearch\NrLlm\Provider\ProviderAdapterRegistryInterface;
 use Netresearch\NrLlm\Service\Analytics\AnalyticsPeriod;
 use Netresearch\NrLlm\Service\LlmConfigurationServiceInterface;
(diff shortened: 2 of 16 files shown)


CybotTM (commented):

### Fund: der Docblock von `OperationCapabilityMap` wiederholt die widerlegte Behauptung

Ich hatte oben schon angemerkt, dass ADR-138 mit einem falschen Satz für `enforce` argumentiert. Beim Lesen des Codes zeigt sich: **derselbe Satz steht im Docblock der neuen Klasse**, und er trägt dort drei Aussagen, von denen nach #676 keine mehr stimmt:

1. > „The seven wizard discoverers write `chat[,vision,tools,streaming,reasoning]` and nothing else"

   Falsch, und zwar schon vor #676: Groq und OpenRouter schrieben ein flaches `['chat']`, Mistral `['chat','tools']` für *jedes* Modell.

2. > „`streaming` is written by four of the seven discoverers (OpenAI, Anthropic, Gemini, Mistral)"

   Nach #676 schreibt **Mistral kein `streaming` mehr** — weder auf dem abgeleiteten Pfad noch im Fallback-Katalog, weil die API es nicht belegt. Gemini leitet es für unbekannte IDs aus `supportedGenerationMethods` ab.

3. `reasoning` existiert nach #676 gar nicht mehr — es ist kein `ModelCapability`, `CapabilitySet` hat es immer verworfen.

Das ist kein Schönheitsfehler. Der Docblock ist die Begründung dafür, welche Operationen **nicht** erzwungen werden; wer ihn liest, entscheidet auf seiner Grundlage über eine Erweiterung der Map.

**Zu tun vor dem Merge, in dieser Reihenfolge:** #676 mergen, dann diese drei Aussagen hier und den entsprechenden Satz in ADR-138 an den dann tatsächlichen Stand anpassen. Ich schreibe den korrigierten Text jetzt nicht hinein — er wäre falsch, falls #676 nicht zuerst landet.

### Geprüft, ohne Fund

- Die `match`-Arme listen jeden `ProviderOperation`-Fall einzeln statt über einen `default`. Ein neuer Enum-Fall wird damit zur Compile-Zeit zur Entscheidung — richtig für eine Tabelle, deren Lücken Absicht sind.
- Die Begründungen für `Translation` (DeepL ist kein `Model`-Record) und `Metadata` (keine Generierung) tragen unabhängig von der Discoverer-Frage.

PR #654, opened 2026-08-08
feat(governance): read-only effective-policy readout

CybotTM:
## Description

Governance lives in `ext_conf_template.txt`, several TCA tables, `be_groups` and three dashboard widgets. There was no place where an operator could see the **effective** state.

Four keys with actual decision content — `privacy.level`, `privacy.retentionDays`, `tools.dataClassEnforcement`, `skills.minTrustLevel` — each with its value and the FQCN of the reader, **read through the same resolvers the runtime uses**. That is the point: the view cannot drift from behaviour, and a functional test pins it by moving `tools.dataClassEnforcement` to `observe` and asserting policy and view move together.

`ToolCallPolicy::enforcing()` is extracted verbatim into `DataClassEnforcementResolver` and stays in `Service\Tool`. `SkillComposerFactory::resolveMinTrustLevel()` becomes public `minTrustLevel()`.

### No apply path — that is the decision, not an omission (ADR-140)

- `ExtensionConfiguration::set()` is `@internal` (`ExtensionConfiguration.php:150`), documented "Set full extension config" (`:165`), and writes the **whole array** via `setLocalConfigurationValueByPath()` (`:166`). No per-key write exists.
- `additional.php` overrides "spoil the `->set()` call" (core docblock `:145-146`), so an apply would report success while the next request serves the old value.

The Install Tool stays where instance-wide keys are set.

**Correction to one half of that argument.** The plan claimed an apply would destroy ADR-115's "default vs chosen" distinction. `storedEnforcement()` says it reads pre-template-merge, but it reads `$GLOBALS['TYPO3_CONF_VARS']['EXTENSIONS']`, and `synchronizeExtConfTemplateWithLocalConfigurationOfAllExtensions()` (`:197-218`) writes the merged template there on any unknown-key read or Install Tool entry. On an install that has entered the Install Tool since the default flipped, the distinction is **already** gone. An apply would guarantee the loss rather than cause it. The decision stands on the other two grounds; the ADR records the weaker claim accurately rather than repeating the stronger one.

### No provenance column

No resolver exposes provenance, and the template sync writes shipped defaults into `settings.php` regardless — the distinction cannot be reconstructed.

### The seven retention overrides are not shown

They all ship as `0` and resolve to the global value by design. Eleven rows, eight of them showing "30", informs nobody. They belong in the administration chapter (#640).

## Module placement

**Overview**, as a third docheader tab. Analytics answers "what did this install spend and do over time" from usage rows; this answers "what does this install allow right now" — a static property, which is what Overview's readiness cards already report. ADR-119 rules out a thirteenth flat module entry.

The plan's cost warning was right: `buildDocHeaderTabMenu()` builds links to real routes, not in-page tabs, so a new action, template, `Modules.php` entries and both XLIFF files were all needed.

## Related Issue

Closes #639

## Type of Change

- [ ] Documentation update

## Checklist

- [x] I have run `composer ci` and all checks pass — unit, PHPStan level 10 (which runs phpat) and cgl re-run independently of the authoring pass: all SUCCESS; the functional classes executed for real. Rector green under an 8.2 resolve, then restored to 8.4 and the suites re-run on that tree.
- [x] I have added an ADR under `Documentation/Adr/` if this changes the public surface — ADR-140

### Falsification criterion for the extraction — met

The extraction is a pure refactor, so the test was: **no behavioural assertion may change.** `git diff Tests/Unit/Service/Tool/ToolCallPolicyTest.php` contains one import, two construction sites wrapped in `new DataClassEnforcementResolver(...)`, and one parameter type. No assertion, expected value, message or test name. Same for the five other construction sites.

### Three things reviewers should look at

1. **`Tests/Architecture/ModuleSeamTest.php` grew its by-name exception list from six to seven.** `EffectivePolicyReadout` must import `DataClassEnforcementResolver` — that crossing *is* the ticket, since asking the runtime's own resolver is what keeps the view honest. It follows the exact precedent of `OverviewReadinessService` (a core-located read model feeding backend Overview), and the rule's docblock demands a recorded split destination, which is there (`→ nr_llm_backend`). Flagged rather than buried: that docblock also says not to grow the list without a reason.
2. **ADR number is 140, while the issue text says 141.** 140 was reserved for a context-classification ticket that was deliberately dropped (deferred in ADR-139), so it is free. A rename plus one toctree line if you prefer 141.
3. **"Unknown" is not reachable from the current resolvers** — all three fail closed and swallow their own read errors, so on a real install every row is answered. It is implemented as a per-row `catch (Throwable)` and tested with a throwing double, proving one bad row does not blank the others. No nullable constructor argument was added to manufacture the state; that would be the `Null*`-dummy shape ADR-120 rejected. The ADR states the limit.

### No screenshot

The running DDEV instance serves `main`, the reference-only worktree, which does not carry this branch. Taking a screenshot of a backend without the tab would have been worse than none.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- A read-only **Governance** tab on the LLM Overview showing the effective
+  value of the four governance keys that carry a decision —
+  `privacy.level`, `privacy.retentionDays`, `tools.dataClassEnforcement`
+  and `skills.minTrustLevel` — together with the FQCN of the resolver each
+  value came from (ADR-140). Every value is read through the same resolver
+  the runtime uses, so the view cannot drift from behaviour: a mistyped
+  `tools.dataClassEnforcement` reads as `enforce` because that is what the
+  gate applies. A resolver that cannot be asked yields "unknown", never a
+  substituted default. There is deliberately no apply path and no
+  provenance column — the Install Tool stays the place where instance-wide
+  keys are set; ADR-140 records why.
+
+### Changed
+
+- `ToolCallPolicy::enforcing()` moved verbatim into the new
+  `Service\Tool\DataClassEnforcementResolver`, which the tool gate and the
+  governance readout both ask. `ToolCallPolicy` no longer takes an optional
+  `ExtensionConfiguration` and instead requires the resolver; behaviour is
+  unchanged, including the fail-closed matrix of ADR-113.
+- `SkillComposerFactory::minTrustLevel()` is public so the readout can show
+  the level the composer is actually built with.
+
 ## [0.26.0] - 2026-08-06
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,13 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   gate applies. A resolver that cannot be asked yields "unknown", never a
   substituted default. There is deliberately no apply path and no
   provenance column — the Install Tool stays the place where instance-wide
-  keys are set; ADR-140 records why.
+  keys are set; ADR-140 records why. Two rows carry more than a value:
+  `tools.dataClassEnforcement = observe` is annotated as applying to
+  built-in tools only, because the gate enforces the trust-zone ceiling for
+  every MCP tool regardless (ADR-115), and each
+  `privacy.retention.<category>` override that deviates from
+  `privacy.retentionDays` gets its own row — the overrides left at the
+  shipped `0` resolve to the global window and stay out.
 
 ### Changed
 
--- a/Classes/Controller/Backend/LlmModuleController.php
+++ b/Classes/Controller/Backend/LlmModuleController.php
@@ -14,6 +14,7 @@ use Netresearch\NrLlm\Domain\Repository\ProviderRepository;
 use Netresearch\NrLlm\Provider\Contract\ProviderInterface;
 use Netresearch\NrLlm\Provider\Exception\ProviderException;
 use Netresearch\NrLlm\Service\Analytics\AnalyticsPeriod;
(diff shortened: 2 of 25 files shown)


CybotTM (commented):

### Die tragende Entscheidung ist die Extraktion, nicht die Anzeige

`DataClassEnforcementResolver` zieht die `enforcing()`-Logik aus `ToolCallPolicy` heraus, und beide lesen sie jetzt aus derselben Quelle. Genau das macht die Anzeige belastbar: Ein Readout, das seine eigene Auflösung mitbringt, driftet vom Laufzeitverhalten ab, sobald einer der beiden Pfade angefasst wird — und die Abweichung fällt niemandem auf, weil beide plausibel aussehen.

Die Fail-Closed-Richtung bleibt beim Verschieben erhalten: Ausnahme → `true`, kein `tools`-Array → `true`, kein String → `true`, und nur ein wörtliches `observe` (nach `trim`/`strtolower`) beobachtet.

### Einer Frage nachgegangen, die ich zunächst für einen Fund hielt

Der Konstruktor von `ToolCallPolicy` ändert sich von `?ExtensionConfiguration $extensionConfiguration = null` zu einem **pflichtigen** `DataClassEnforcementResolver`. Das sieht nach einem Breaking Change aus, und im CHANGELOG steht keiner.

Ist keiner, nach der Konvention dieses Repos: `api-surface.txt` führt nur `ToolCallPolicyInterface` — die konkrete Klasse steht nicht drin — und `ApiSurfaceSnapshotTest` schließt `__construct` ausdrücklich aus („service constructors are DI wiring and out of scope"). Das Interface bleibt unverändert.

### Geprüft, ohne Fund

- Jede Ablesung ist eingepackt; ein werfender Resolver liefert eine „unbekannt"-Zeile statt eines Wertes. Für eine Governance-Ansicht ist das die richtige Richtung — ein Operator handelt auf der Anzeige.
- Nur *abweichende* Retention-Overrides bekommen eine Zeile, nicht die sieben, die als `0` ausgeliefert werden und auf das globale Fenster auflösen.
- Kein Apply-Pfad, mit Begründung im ADR.

PR #676, opened 2026-08-09
fix(discovery): seed the capabilities the provider reports, not a guess

CybotTM:
## Description

Model discovery wrote capability tokens no provider response substantiated. Six of the seven discoverers were wrong, and the seeded values are what a criteria-mode configuration matches against — so the wrong data selects the wrong model, or none.

### What each provider actually reports

Verified against the live endpoints and the vendors' own API references, not inferred:

| Discoverer | wrote | reports | now reads |
|---|---|---|---|
| Mistral | `chat, tools` for every model | per-model `capabilities` object | `completion_chat`, `function_calling`, `vision` |
| OpenRouter | flat `chat` | `supported_parameters`, `architecture.input_modalities` | `tools`, image input |
| Ollama | `tools` if the tag contains `qwen`, `llama3`, `mistral` or `mixtral` | `/api/show` → `capabilities` | the reported array |
| Gemini | `chat, vision` for any id outside its table | `supportedGenerationMethods` | `generateContent`, `streamGenerateContent`, `embedContent` |
| Groq | `chat` | nothing — the listing has no capability field | unchanged, and the docs say why |
| OpenAI | id table, `chat` for unknown ids | — | unchanged |
| Anthropic | `chat, vision, tools, streaming` | — | unchanged; the listing returns Claude chat models only |

The Ollama guess was wrong in both directions: a tool-capable model outside those four families never got the token, and any distilled or quantised variant carrying one of those words in its tag got it without being able to call a tool.

### A second defect, found by checking the live API

`OpenRouterProvider::fetchModels()` reported **no** capabilities for **any** model:

- it read `supports_function_calling`, which is not a field the catalogue returns;
- it compared `architecture.modality` against `"multimodal"`, which is not a value that field takes — it is a display string such as `text+image->text`.

Measured against the live catalogue: 400 models, 0 carrying either signal, against 333 that list `tools` among their supported parameters and 237 that accept image input.

This survived 109 green tests because the fixtures used the same invented shape. The fixtures now carry the real one; the assertions are unchanged.

### `reasoning` is gone

Two discoverers wrote it. It is not a `ModelCapability`, so `CapabilitySet` discarded it and the TCA checkbox list could not display it — a value in the stored CSV that nothing read and nobody could edit.

### The decision behind "narrow rather than guess"

`capabilities` is an operator-editable checkbox list in the model record. The field is theirs; discovery seeds it. So a discoverer states what the API states, and where the API is silent (Groq) it makes the narrowest true claim instead of a plausible one. The administration chapter now says per provider how much is seeded and asks the operator to complete it.

**This bears on #653.** A capability check with `enforce` as its default judges an operator-owned field that the wizard only pre-fills. On any installation that never revisited it — every installation before this change — the check refuses models that work. Worth deciding on #653 itself, not here.

## Related Issue

Closes #671. Prerequisite for #653 / #635.

## Type of Change

## Migration

Existing `tx_nrllm_model` rows keep their values: an upgrade cannot tell an operator's deliberate edit from a stale seed, so it overwrites neither. Re-run **Fetch Models**, or correct the checkboxes by hand. Documented in the administration chapter.

## Checklist

- [x] I have run the gates — unit, PHPStan (level 10), cgl, fuzzy, functional and Rector (under an 8.2 resolve, as CI pins it)
- [ ] ADR — none: this corrects data the existing ADR-018 discovery already promised, it does not change the surface

### Tests

`CapabilitySeedTest` pins each discoverer against a trimmed recording of the real endpoint, plus a vocabulary test asserting no discoverer can emit a token outside `ModelCapability`.

Controlled before/after rather than plausibility: against the unchanged `Classes/`, 5 of its 7 tests fail. The two that pass in both are characterisation of behaviour deliberately left alone (Groq, and Ollama below 0.6).

Three existing tests asserted the old guesses and were rewritten rather than deleted: the two Ollama capability tests now source the capability from the server's report instead of the tag, keeping their assertions; the OpenAI spec test lost `reasoning` from its expectations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,36 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- Model discovery seeds capabilities the provider actually reports (#671).
+  Six of the seven discoverers wrote tokens no response substantiated:
+  Groq and OpenRouter a flat `chat`, Mistral `chat, tools` for every model
+  including vision-only ones, Gemini `chat, vision` for any id outside its
+  built-in table, and Ollama `tools` for any tag containing `qwen`,
+  `llama3`, `mistral` or `mixtral` — a guess from the model NAME that was
+  wrong in both directions. Mistral's per-model `capabilities` object,
+  OpenRouter's `supported_parameters` and `architecture.input_modalities`,
+  Ollama's `/api/show` `capabilities` array and Gemini's
+  `supportedGenerationMethods` are read instead. Groq's listing carries no
+  capability field, so its models stay `chat` and the administration
+  chapter says so.
+  This matters beyond the wizard: `Model::$capabilities` began reaching the
+  entity in 0.26, so configurations selecting a model by criteria match
+  against these values for the first time.
+  Also dropped: the `reasoning` token, written by two discoverers, absent
+  from `ModelCapability` — `CapabilitySet` discarded it and the TCA
+  checkbox list could not show it.
+  Existing `tx_nrllm_model` rows keep their values; re-run **Fetch Models**
+  to refresh them.
+- `OpenRouterProvider::fetchModels()` reported no capabilities for any
+  model. It read `supports_function_calling`, which is not a field the
+  catalogue returns, and compared `architecture.modality` against
+  `"multimodal"`, which is not a value it takes — the field is a display
+  string such as `text+image->text`. Measured against the live catalogue:
+  400 models, 0 with either signal, against 333 that list `tools` among
+  their supported parameters and 237 that accept image input.
+
 ## [0.26.0] - 2026-08-06
 
 ### Added
--- a/Classes/Provider/OpenRouterProvider.php
+++ b/Classes/Provider/OpenRouterProvider.php
@@ -248,8 +248,16 @@ final class OpenRouterProvider extends AbstractProvider implements
                         'completion' => $this->asFloat($pricing['completion'] ?? 0),
                     ],
                     'capabilities' => [
-                        'vision' => $this->getString($architecture, 'modality') === 'multimodal',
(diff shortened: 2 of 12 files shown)


CybotTM (commented):

Kontingentlimit; die COMMENTED-Zeile ist ein Fehler, keine Bewertung. Als Autor gelesen. Der PR-Text dokumentiert die Verifikation gegen die Live-Endpunkte; hier steht nur, was ich beim zweiten Durchgang zusätzlich geprüft habe.

### Randfälle nachgerechnet

- **Mistral, `capabilities` vorhanden aber ohne `completion_chat`** → `?? true` greift, das Modell bekommt `chat`. Richtige Richtung: die Auflistung ist eine Chat-Modellliste.
- **Mistral-Embedding-Modell** (`completion_chat: false`, keine weiteren Flags) → **leere** Capability-Liste. Das ist beabsichtigt und ehrlich, aber es lohnt, es auszusprechen: Ein solcher Datensatz matcht danach *kein* Kriterium mehr. Eine Regression ist es nicht — vorher stand dort `chat,tools`, was ein `embeddings`-Kriterium genauso wenig erfüllt hat. `embeddings` ließe sich aus `completion_chat === false` nur *raten*, und Raten ist das, was dieser PR entfernt.
- **Ollama meldet `capabilities: []`** (leeres Array statt fehlendem Schlüssel) → fällt auf `['chat']`, nicht auf eine leere Liste. Der `=== []`-Zweig fängt beides.
- **OpenRouter ohne `architecture`** → `input_modalities` leer, kein `vision`, `chat` bleibt. Kein Notice.

### Zur Reihenfolge

Der PR ist Vorbedingung für #653, und #653s `OperationCapabilityMap` trägt im Docblock noch die widerlegte Behauptung über die Discoverer — inklusive zweier Aussagen, die erst *durch* diesen PR falsch werden (Mistral schreibt kein `streaming` mehr, `reasoning` gibt es nicht mehr). Dort kommentiert.

PR #678, opened 2026-08-09
fix(tools): gate FAL files on the acting user's mounts, not the ambient one's

CybotTM:
## Description

`FalStorageGate::isFileAccessible()` evaluated two halves against two different people:

- the allow-list (`isAllowed()`) and the admin shortcut used the **passed** user;
- the mount and permission half used whoever was in `$GLOBALS['BE_USER']`.

Harmless today: the three read-only FAL tools run synchronously inside their owner's request, so both users are the same person. It stops being harmless on the approval path, where a run resumes under its **owner's** identity (ADR-083) while the **approver** is ambient. A broader-mounted approver authorises a file the owner cannot reach; a narrower one refuses a legitimate read. The docblock claimed the opposite.

## Two contaminated caches, not one

This is the part the issue did not have, and it changes the fix.

1. `StorageRepository::findByUid()` returns the request-shared instance whose mounts and permissions the core `StoragePermissionsAspect` attached **once**, for the ambient user at creation time. `ResourceStorage` keeps no reference to the global afterwards. So the issue's suggested direction — `BackendUserAuthentication::getFileStorages()` — resolves through the same repository and returns the same objects. It would have changed nothing.

2. `BackendUserAuthentication::getFileMountRecords()` memoises into the runtime cache under the single key `backendUserAuthenticationFileMountRecords`, with no user in it. The first backend user to ask in a request populates it and every later user object receives **that** user's mounts.

Two users in one request is exactly the situation this gate exists for, so both public routes answer for the wrong person.

## What it does now

- builds its own storage from the `sys_file_storage` record via `StorageRepository::createFromRecord()`, which dispatches no `AfterResourceStorageInitializationEvent` and never enters the instance cache;
- applies the passed user's file permissions and `permissions.file.storage.<uid>` TSconfig, mirroring what the aspect computes;
- reads the `sys_filemounts` rows for that user's own mount uids.

`groupData` is core-`@internal` and used deliberately: it is the only per-user source. If it ever disappears, the mount list is empty, the storage gets no mounts and the gate denies — loudly, and closed. Preferred over depending on an undocumented cache key, where a rename would make the fix a silent no-op.

Per-user storages are memoised for the gate's lifetime because two of the three consumers call this once per result row, and each miss builds a driver. That is the only reason the class is no longer `readonly`.

## Related Issue

Closes #672.

## Type of Change

## Checklist

- [ ] ADR — none: this makes the code do what ADR-047 and the class docblock already said

### Tests

`FalStorageGateActingUserTest` is the acceptance test the issue asked for: two editors with **disjoint** mounts on the same storage, so the storage allow-list cannot tell them apart and only the mount check can. Each test logs one in and asks the gate about the other, in both directions and with the roles swapped.

Against unchanged `Classes/`, three of its four tests fail. The fourth — one user in both roles — passes either way by design: without it, a gate that simply denied everything would satisfy half the assertions.

The test also forces the shared storage into existence under the ambient user before asking, so the defect depends on the bug rather than on call order.

Worth recording: the first attempt at this fix looked right and still left three assertions red. That is how the runtime-cache contamination surfaced — the test was written before the fix, not after it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- `FalStorageGate::isFileAccessible()` enforces the mounts of the user it is
+  passed, not of whoever is in `$GLOBALS['BE_USER']` (#672). The storage
+  allow-list half already used the passed user; the mount half did not, and
+  the two disagreed whenever the acting and ambient user differ — which is
+  exactly the approval path, where a run resumes under its owner's identity
+  (ADR-083) while the approver is ambient. A broader-mounted approver
+  authorised a file the owner could not reach; a narrower one refused a
+  legitimate read.
+  Two request-shared caches caused it, and neither is reachable through a
+  per-user API: `StorageRepository::findByUid()` returns the cached
+  `ResourceStorage` whose mounts the core `StoragePermissionsAspect`
+  attached once for the ambient user, and
+  `BackendUserAuthentication::getFileMountRecords()` memoises into the
+  runtime cache under one key with no user in it — so the first user to ask
+  in a request answers for every later one. The gate now builds its own
+  storage from the record via `createFromRecord()` (which dispatches no
+  initialization event and never enters the instance cache) and reads the
+  mount rows for that user's own mount uids.
+  Behaviour is unchanged for the three read-only FAL tools shipped today:
+  they run synchronously in their owner's request, where both users are the
+  same person.
+
 ## [0.26.0] - 2026-08-06
 
 ### Added
--- a/Classes/Service/Tool/FalStorageGate.php
+++ b/Classes/Service/Tool/FalStorageGate.php
@@ -219,13 +219,9 @@ final class FalStorageGate
      */
     private function fileMountRecordsFor(BackendUserAuthentication $user): array
     {
-        if (!$this->connectionPool instanceof ConnectionPool) {
-            return [];
-        }
-
         $declared  = $user->groupData['filemounts'] ?? '';
         $mountUids = is_scalar($declared) ? GeneralUtility::intExplode(',', (string)$declared, true) : [];
-        if ($mountUids === []) {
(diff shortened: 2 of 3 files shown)


CybotTM (commented):

### Randfälle nachgerechnet

- **Cache-Schlüssel `$user->getUserId() . ':' . $storageUid`.** `getUserId()` ist `?int` — bei `null` entstünde der Schlüssel `":1"`, den sich zwei Benutzer teilen könnten. Unerreichbar: `isFileAccessible()` ruft zuerst `isAllowed()`, das über `getFileStorages()` geht; ein Benutzerobjekt ohne uid hat dort keine Storages, die Prüfung endet vorher. Ich lasse es, weil ein zusätzlicher Guard einen Pfad absichern würde, den es nicht gibt.
- **`fileMountRecordsFor()` bei leerem `groupData['filemounts']`** → leere Liste → Storage ohne Mounts → `checkFileActionPermission` verweigert. Fail-closed, und genau die Richtung, die der Docblock für den Fall verspricht, dass `groupData` einmal verschwindet.
- **Admin.** Kehrt vor der ganzen Konstruktion zurück; die teure Storage-Erzeugung entfällt für den häufigsten Fall.
- **`createFromRecord()` landet nicht im Instanz-Cache** — geprüft an `StorageRepository`: das Init-Event wird ausschließlich in `getStorageObject()` ausgelöst, nicht in `createStorageObject()`. Deshalb hängt sich der Aspect nicht an unser Objekt, und unser Objekt hängt sich nicht in den geteilten Cache.

### Was ich hervorheben würde

Der interessanteste Teil ist nicht der Fix, sondern dass der *naheliegende* Fix nicht funktioniert hätte. Der im Issue vorgeschlagene Weg über `getFileStorages()` löst durch dieselbe Repository-Instanz auf und hätte nichts geändert; `getFileMountRecords()` memoisiert unter einem Schlüssel ohne Benutzer. Beide öffentlichen Wege antworten für den ambienten Benutzer. Aufgefallen ist es nur, weil der Test vor dem Fix geschrieben war und drei Assertions rot blieben, obwohl der Code richtig aussah.

Kein weiterer Fund.

PR #679, opened 2026-08-09
fix(tools): pin the live workspace where the FAL tools read metadata

CybotTM:
## Description

`sys_file_metadata` carries `'versioningWS' => true`, so a file with a draft version has more than one row for the same `file` uid. Neither of the two tools that read it said which row it wanted.

**`read_fal_asset_meta`** returned whichever row the database happened to give — an unpublished draft as readily as the live value — and reported it as current. A model reading it gets a value no visitor sees; a human reading the transcript cannot tell. Now pinned with `WorkspaceRestriction(0)`, `ORDER BY uid ASC` and `setMaxResults(1)`: the same three things core's own `MetaDataRepository::findByFileUid()` pins for the identical lookup.

## The sweep found a second one

The issue asked whether any other builtin reads a workspace-aware table unrestricted. **`search_fal_files`** does, and the consequence is worse than a wrong field:

- a search for a word that appears **only in an unpublished draft** returns the file;
- a file with a draft version is **listed twice**, because both rows join.

Measured before fixing, not argued from the query — the two new tests were red at `Unpublished` matching and at `2 is identical to 1`. The join now requires `t3ver_wsid = 0`, spelled out rather than added as a `WorkspaceRestriction` because the restriction applies to the query's own FROM table, not to a join alias.

## Deliberately not touched

`GetTsConfigTool` and `GetRecordHistoryTool` also query the workspace-aware `pages` table without a restriction, and both are staying as they are: they look up `uid = <the value the caller passed>`, which returns exactly the row that was asked for. There is no arbitrary-row problem. Whether a tool should accept a *version* uid at all is a different question, and not this fix's.

The remaining `removeAll()` readers touch `sys_file`, `be_users`, `be_groups` and `tx_scheduler_task`, none of which are workspace-aware.

## Related Issue

Closes #674.

## Type of Change

## Checklist

- [ ] ADR — none: this makes two queries do what the surrounding contract already assumed

### Tests

The acceptance test the issue asked for: a live row and a draft version whose uid sorts **before** it, so an unpinned query prefers the draft. Red without the fix, green with it.

Two more for the search tool: draft-only text must not match, and a file with a draft version must be listed once. Both red before the fix.

### Note

Both defects are read-only — the answer could be wrong, nothing was corrupted. The same unpinned query in the *writing* tool is fixed on the `set_file_alternative_text` branch, where a wrong write is reported as success.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- The two FAL tools that read `sys_file_metadata` pin the live workspace
+  (#674). The table carries `'versioningWS' => true`, so a file with a draft
+  version has more than one row for the same `file` uid:
+  - `read_fal_asset_meta` returned an arbitrary one — potentially an
+    unpublished draft — as the current value, with nothing in the answer
+    saying so. It now adds `WorkspaceRestriction(0)`, `ORDER BY uid ASC` and
+    `setMaxResults(1)`, the same three things core's own
+    `MetaDataRepository::findByFileUid()` pins.
+  - `search_fal_files` joined the table without a workspace condition, so a
+    search for text that exists only in a draft returned the file, and a file
+    with a draft version was listed twice. The join now requires
+    `t3ver_wsid = 0`.
+
+  Both defects are read-only: the answer could be wrong, nothing was
+  corrupted. The same defect in the writing tool is fixed separately in the
+  `set_file_alternative_text` branch, where it mattered more.
+
 ## [0.26.0] - 2026-08-06
 
 ### Added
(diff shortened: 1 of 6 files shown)


CybotTM (commented):

### Geprüft

- **`t3ver_wsid = 0` in der JOIN-Bedingung statt als `WorkspaceRestriction`.** Richtig so: die Restriction wirkt auf die FROM-Tabelle der Abfrage, nicht auf einen Join-Alias. Ein `WorkspaceRestriction` an dieser Stelle hätte `sys_file` eingeschränkt — eine Tabelle, die gar nicht workspace-fähig ist — und die Metadaten unberührt gelassen.
- **`ORDER BY uid ASC` plus `setMaxResults(1)`** auch dort, wo die Restriction bereits genügen sollte. Doppelt gemoppelt, aber die Begründung im Kommentar trägt: zwei Live-Zeilen für dieselbe Datei sollte es nicht geben, und falls doch, ist eine *stabile* Antwort besser als eine wechselnde — und es ist dieselbe Zeile, die das schreibende Tool trifft.
- **Der Sweep** ist im PR-Text nach beiden Seiten abgegrenzt: was gefunden wurde (`search_fal_files`) und was bewusst unberührt bleibt (`GetTsConfigTool`, `GetRecordHistoryTool`, weil sie auf `uid = <übergebener Wert>` nachschlagen).

Beide neuen Tests waren vor dem Fix rot, einer davon mit `2 is identical to 1` — die Dublette war also real und nicht nur theoretisch.

Kein Fund.

PR #642, opened 2026-08-08
feat(tools): a builtin's declared write effect implies human approval

CybotTM:
## Description

Two declarations describe a tool that changes something, and they were unconnected.

`ToolEffectInterface` (ADR-111) feeds the write fence, the lease, the audit step and the retry decision — four readers, none about authorisation. `RequiresApprovalInterface` (ADR-084) is the only thing the approval scan looks for, and **no production class implements it**: the sole implementers are two anonymous test classes. A tool could therefore declare `NON_IDEMPOTENT_WRITE` and still run unattended.

The approval scan in `ToolLoopService` now also binds on a declared write. The effect is the better of the two to key on — it is a property of the code and deliberately not configurable (ADR-111), so an administrator cannot relabel a write as a read to dodge it.

### The remote exemption is load-bearing

`McpTool::getEffect()` returns `NON_IDEMPOTENT_WRITE` for **every** imported tool, a pure search included. That value is a fail-closed assumption about a body this codebase cannot inspect (ADR-116), not the tool's statement about itself. Treating it as one would suspend every MCP tool on every call and leave the shipped MCP client unusable. `readOnlyHint` is deliberately not consulted — a remote server must not influence its own authorisation. The remote axis gets an operator-declared, server-level column separately.

Precedence: an explicit marker wins over the remote exemption.

### The registration ban is extended

`ToolRegistry` already refuses a tool implementing both `RequiresApprovalInterface` and `RequiresInputInterface` (ADR-105). Since a declared write is now a second way to be approval-bound, the ban extends to it: a **non-remote, write-declaring** tool may not implement `RequiresInputInterface` either.

Not defensive tidying — without it the combination is dead rather than handled. The approval scan runs *before* the input scan, so such a tool suspends `AWAITING_APPROVAL` and never `WAITING_FOR_INPUT`; `resume()` then refuses it for the input it never received, the model re-requests, and it suspends again. One operator decision per cycle, `submitInput()` unreachable. Failing at container boot beats failing silently at runtime.

The predicate mirrors `ToolLoopService::requiresHumanApproval()`, remote exemption included — narrowing the exemption in one place means narrowing it in the other.

> **Korrektur (2026-08-09):** Diese Passage stand vorher als „Not done, on purpose" hier und behauptete das Gegenteil. Sie war veraltet — ADR-134 und der Code haben die Ausweitung von Anfang an, nur der PR-Text nicht. Aufgefallen beim manuellen Review, weil Copilots Kontingent erschöpft ist.

[...]

PR #680, opened 2026-08-09
test(agent): make the deferred input-path gate self-enforcing

CybotTM:
## Description

`ResumeCoordinator::submitInput()` authorises the submitter with `agent_approve` and nothing else — it never checks them against the tool whose input they supply — while `ToolLoopService::resumeWithInput()` executes the pending calls under the **run owner's** context (ADR-083). A non-admin holding `agent_approve` could therefore satisfy an admin-only tool that then runs on the owner's authority: the confused deputy #622 closed on the approval side.

It is latent, not live. **No production class implements `RequiresInputInterface`** — the only references are the loop's own `instanceof` checks and the registration ban in `ToolRegistry`. Nothing can suspend for input at all.

## No gate is built here, on purpose

#622's fix cannot be copied, and the difference is not a detail:

- **No write axis.** #622 gates the calls that WRITE, because a write is what must not happen unattended. An input-requiring tool declares no effect — ADR-105 makes the two markers mutually exclusive at registration — so there is no "which calls matter" axis. Gating *every* input submission is a broader policy and a different decision.
- **No digest.** The approval path binds a decision to the reviewed turn (ADR-132). The input path has no equivalent, so a submission cannot be shown to belong to the turn it was written for.

Choosing an axis and a binding now would be guessing at a policy for a tool nobody has proposed — which is what ADR-120 exists to prevent. The issue says the same thing in its own suggested direction.

## What ships instead

The tripwire, in the shape `ToolEffectCoverageTest` already established for the write side: the empty set is pinned, so the first tool that can suspend for input fails the build. Its author then owns the two questions, named in the docblock — on what axis the submitter is gated against the tool, and whether the input path gets a turn digest of its own.

A note in an issue nobody re-reads becomes a red build at exactly the moment the deferral stops being safe.

## Related Issue

Closes #649.

## Type of Change

- [x] New feature (non-breaking change that adds functionality) — a test, not behaviour

## Checklist

- [ ] ADR — none: this records a deferral, it does not decide anything

### Verification

A test that can only ever pass is worthless, so it was made to fire rather than assumed to work: giving `ReadFalAssetMetaTool` the interface turned the test red and named the tool in the failure message. Reverted after the probe — `git diff` on `Classes/` is empty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,20 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- `InputPauseCoverageTest` pins that no builtin can suspend a run for
+  operator input (#649). The input path authorises the submitter with
+  `agent_approve` alone and never against the tool whose input they supply,
+  while the resume executes under the run owner's context (ADR-083) — the
+  confused deputy the approval path closed in #622. It is unreachable only
+  because nothing implements `RequiresInputInterface`, so the first tool
+  that does now turns a latent gap into a red build. The gate itself is
+  deliberately not built ahead of that tool: #622's cannot be copied,
+  because an input-requiring tool declares no effect to gate on (ADR-105
+  makes the two markers mutually exclusive) and the input path has no turn
+  digest (ADR-132).
+
 ## [0.26.0] - 2026-08-06
 
 ### Added
--- a/Tests/Functional/Service/Tool/InputPauseCoverageTest.php
+++ b/Tests/Functional/Service/Tool/InputPauseCoverageTest.php
@@ -1,0 +1,98 @@
+<?php
+
+/*
+ * Copyright (c) 2025-2026 Netresearch DTT GmbH
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+declare(strict_types=1);
+
+namespace Netresearch\NrLlm\Tests\Functional\Service\Tool;
+
+use Netresearch\NrLlm\Service\Tool\RequiresInputInterface;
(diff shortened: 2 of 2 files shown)


CybotTM (commented):

### Die Frage bei diesem PR ist nicht der Code, sondern ob das Nicht-Bauen richtig ist

Ich habe die Begründung gegen den Code geprüft, nicht gegen das Issue:

- **Keine Schreib-Achse.** `ToolRegistry` verbietet `RequiresApprovalInterface` + `RequiresInputInterface`, und #642 weitet das auf effektdeklarierende Tools aus. Ein eingabeforderndes Tool kann also gar keinen Schreibeffekt deklarieren — die Achse, auf der #622 gattert, existiert hier nicht. Trägt.
- **Kein Digest.** `PendingTurnDigest` wird ausschließlich auf dem Approval-Pfad gebildet und geprüft; `submitInput()` kennt nichts Vergleichbares. Trägt ebenfalls.

Beides sind Eigenschaften des Codes, nicht Meinungen — ein Gate jetzt hätte eine Achse erfinden müssen.

### Am Test selbst

`builtinNames()` hydriert bewusst **nicht** die Provider, ist also der Compile-Zeit-Satz. Damit kann ein von außen geliefertes Tool die Zusicherung weder erfüllen noch brechen — dieselbe Wahl wie in `ToolEffectCoverageTest`. Die `assertGreaterThan(0, $checked)`-Zeile bewacht den Wächter: eine leer gewordene Registry würde die Hauptbehauptung sonst leer erfüllen.

Die Kontrollprobe (Interface an `ReadFalAssetMetaTool` gehängt → Test rot, Toolname in der Meldung → zurückgesetzt) ist der Teil, ohne den ein solcher Tripwire wertlos wäre.

Kein Fund.

PR #681, opened 2026-08-09
fix(config): state the condition on the enforcement pin instead of promising it

CybotTM:
## Description

The Install Tool label for `tools.dataClassEnforcement` stated, without qualification:

> An install upgraded from an older version **is pinned to Observe by an upgrade wizard** so the change does not silently strip tools from a working setup

It usually is not, and the mechanism is worth spelling out because the two halves are in different components:

- `DataClassEnforcementDefaultUpdateWizard::updateNecessary()` is `hasProviders() && storedEnforcement() === null`, and `storedEnforcement()` reads `$GLOBALS['TYPO3_CONF_VARS']['EXTENSIONS']['nr_llm']['tools']['dataClassEnforcement']`.
- `ExtensionConfiguration::synchronizeExtConfTemplateWithLocalConfigurationOfAllExtensions()` writes the merged template into exactly that place. It runs on entering the Install Tool, on `extension:setup`, and from `ExtensionConfiguration::get()` when the key is absent.
- `ext_conf_template.txt` ships `enforce`.

So after any of those the value is no longer `null`, the wizard reports nothing to do, and the operator lands on **enforce** — with the gate removing tools since the update, and the label telling them the opposite in the screen where they would look.

This is not a stale sentence in a document someone may not open. It is the field description shown while deciding what to set, asserting a protection that may never have happened on their install.

## What changed

The label states the condition instead of the outcome: the wizard offers the pin only while the field has never been written, and after an upgrade the value shown is what applies — whether or not a wizard ran.

No behaviour change. Whether the wizard *should* pin more reliably is a separate question; this PR makes the screen stop claiming it already does.

## Related Issue

Closes #675.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue) — a wrong statement in the operator-facing configuration surface

## Checklist

- [ ] ADR — none: ADR-113/115 describe the intent, which is unchanged

### Tests

A characterisation test pins the limit the label now describes: with a provider present and the **template default** stored, `updateNecessary()` is false. The existing `respectsAnExplicitEnforceChoice` covers the same code path but reads as though the value were always an operator's decision — the new test says in its name and message that the wizard cannot tell the two apart, because both are the same string in the same place.


CybotTM (commented):

### Die Kette nachgeprüft, nicht übernommen

Die Behauptung des PRs ist eine über drei Komponenten hinweg, und genau solche Ketten stimmen oft nur zur Hälfte:

1. `updateNecessary()` = `hasProviders() && storedEnforcement() === null` — gelesen, stimmt.
2. `storedEnforcement()` liest `$GLOBALS['TYPO3_CONF_VARS']['EXTENSIONS']['nr_llm']['tools']['dataClassEnforcement']` — gelesen, stimmt.
3. `synchronizeExtConfTemplateWithLocalConfigurationOfAllExtensions()` schreibt genau dorthin, und `ext_conf_template.txt` liefert `enforce` aus.

Damit ist der Wert nach jedem Install-Tool-Besuch gesetzt und der Wizard meldet nichts zu tun. Die Kette trägt.

### Zum Test

Der neue Test dupliziert scheinbar `respectsAnExplicitEnforceChoice` — derselbe Codepfad, dasselbe Ergebnis. Der Unterschied liegt in Name und Meldung: Der bestehende Test liest sich, als sei der gespeicherte Wert immer eine Operator-Entscheidung. Der neue sagt, dass der Wizard das gar nicht unterscheiden kann, weil beide derselbe String an derselben Stelle sind. Das ist der Punkt, den das Label vorher falsch dargestellt hat.

Kein Fund.

PR #682, opened 2026-08-09
test(ci): run the three PHP tests that were in no suite

CybotTM:
## Description

`Tests/E2E/TCA/` was in neither suite of `Build/FunctionalTests.xml`. The issue asked whether anything else lives under `Tests/E2E/` outside `Backend/` — it does, and it is worse than expected:

| Test | was in | ran in |
|---|---|---|
| `Tests/E2E/TCA/TcaFieldCompletionTest` | no suite of `FunctionalTests.xml` | nothing |
| `Tests/E2E/ChatCompletionWorkflowTest` | `phpunit.xml`'s `e2e` suite | nothing |
| `Tests/E2E/EmbeddingWorkflowTest` | `phpunit.xml`'s `e2e` suite | nothing |

The third layer: `Build/phpunit.xml` **does** declare an `e2e` testsuite over `../Tests/E2E`, but no `runTests.sh` selector invokes it — `-s e2e` runs Playwright. A declaration with no reader.

Measured rather than assumed: all three answered `No tests executed` before this change, and run now (15 and 11 tests).

That is why a failing assertion could sit in `TcaFieldCompletionTest` unnoticed. The test is not weak; it was never executed. **A test no job runs is worse than no test, because it reads as coverage.**

## The label finding

The assertion it failed on is right, and so is the TCA. `tx_nrllm_configuration.preset_checksum` is `type => passthrough`, written by `ConfigurationPresetImportService` and — per the comment above it — never edited in FormEngine. It appears in no `showitem`. A label would itself be a declaration nothing reads, so the field is not given one.

Passthrough columns are now skipped **by their TCA type, not by name**, so the rule holds for the next such column instead of growing an exemption list.

## Where the tests went

- `Tests/E2E/TCA/` → a third `e2e-tca` suite in `FunctionalTests.xml`, which is the config CI runs.
- The two workflow tests extend `AbstractUnitTestCase` and use mocked providers, so they are unit tests that happen to live under `Tests/E2E/`. Listed as explicit `<file>` entries in the `unit` suite — a `<directory>` there would also pull in `Tests/E2E/Backend`, which belongs to the functional config.

The parallel functional runner globs its directories with `find`, so `Tests/E2E/TCA` is added there too. Without it the new suite would be silently skipped in that mode — exactly what the comment above that line already warns about.

## Related Issue

Closes #658.

## Type of Change

## Checklist

- [ ] ADR — none: this runs tests that already existed

### Note for reviewers

The `AGENTS.md` line naming the PHPUnit suites sits inside an `AGENTS-GENERATED` block. There is no generator in the repo to update instead, and the line was factually wrong without the edit — worth a look if that block is regenerated from somewhere outside this repository.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -60,7 +60,7 @@ ALWAYS use the Docker test runner; never invoke `phpunit` / `phpstan` / `rector`
 ## Testing
 
 - Unit / Integration / Fuzzy / Functional / E2E suites — see `Tests/AGENTS.md` for layout details.
-- PHPUnit configs: `Build/phpunit.xml` (unit + integration + fuzzy), `Build/FunctionalTests.xml` (functional + e2e-backend).
+- PHPUnit configs: `Build/phpunit.xml` (unit + integration + fuzzy; the unit suite also lists the two workflow tests under `Tests/E2E/`), `Build/FunctionalTests.xml` (functional + e2e-backend + e2e-tca). Every PHP test must sit in one of these suites — a directory in none of them runs in no job and reads as coverage (#658).
 - Mutation: `infection.json.dist` (target MSI ≥ 70%).
 - Architecture tests: `Tests/Architecture/` (phpat) — enforce layered boundaries (Controller → Service → Provider).
 <!-- AGENTS-GENERATED:END testing -->
--- a/Build/FunctionalTests.xml
+++ b/Build/FunctionalTests.xml
@@ -21,6 +21,15 @@
         <testsuite name="e2e-backend">
             <directory>../Tests/E2E/Backend/</directory>
         </testsuite>
+        <!--
+            Tests/E2E/TCA holds plain PHPUnit tests over the TCA files. They
+            were in no suite at all and therefore ran in no job: a failing
+            assertion sat there unnoticed because nothing executed it (#658).
+            Listed here because this config is the one CI runs.
+        -->
+        <testsuite name="e2e-tca">
+            <directory>../Tests/E2E/TCA/</directory>
+        </testsuite>
     </testsuites>
     <!--
         Without <source> PHPUnit answers any coverage request with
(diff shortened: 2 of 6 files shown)


CybotTM (commented):

### Geprüft

- **Beide Eintragsorte.** `FunctionalTests.xml` bekommt die dritte Suite, und der Parallel-Pfad in `runTests.sh` globt seine Verzeichnisse separat per `find` — ohne den zweiten Eintrag wäre die neue Suite im Parallelmodus still übersprungen worden. Der Kommentar über genau dieser Zeile warnt davor; er ist mitgeändert, sonst zählt er weiter zwei Verzeichnisse auf.
- **`<file>` statt `<directory>`** für die zwei Workflow-Tests in der Unit-Suite. Ein `<directory>` auf `../Tests/E2E` hätte `Backend/` mitgezogen, das zur funktionalen Konfiguration gehört und dort doppelt liefe.
- **Die Passthrough-Ausnahme nach Typ, nicht nach Name.** `preset_checksum` steht in keinem `showitem` und ist per Kommentar „never edited in FormEngine" — ein Label wäre selbst eine Deklaration ohne Leser. Nach `config.type === 'passthrough'` auszunehmen hält die Regel für die nächste solche Spalte, statt eine Liste wachsen zu lassen.
- **Kontrolle vorher/nachher**: `No tests executed` für beide Selektoren vor der Änderung, 15 bzw. 11 Tests danach. Ohne diese Messung wäre „läuft jetzt" eine Behauptung.

### Ein Hinweis für Reviewer

Die geänderte Zeile in `AGENTS.md` steht in einem `AGENTS-GENERATED`-Block. Im Repo gibt es keinen Generator, der sie erzeugt — wer den Block von außen neu generiert, überschreibt die Korrektur. Das steht auch im PR-Text.

Kein Fund.

PR #683, opened 2026-08-09
fix(skills): place the block on a multimodal user turn, and stop losing it silently

CybotTM:
## Description

`SkillInjectionService::isUserMessage()` recognised an array-shaped message as a user message only when its `content` was a **string**. `augmentMessages()` prepends to the *first* message satisfying that predicate, so a conversation whose first user turn is multimodal had the block prepended to some later text turn — or, with no later text turn, not at all. Silently, either way.

Multimodal content is a supported input shape, not an accident: `MessageShaper::normalise()` converts only the exact 2-key string/string form into a `ChatMessage` and passes everything else through, and its docblock names multimodal messages as the reason. The adapters expect to receive them.

## Both halves, not one

The issue offers two directions and says the "no injection site" branch should stop being silent whichever is chosen. Both are here, because they close different things:

**The behaviour.** The block is prepended as a leading `{type: 'text'}` part — the OpenAI-style shape every adapter reads, which `ClaudeProvider::convertMultimodalContent()` translates into Claude's own format. Existing parts keep their order and are never merged into, because a part may be an image.

An associative content array is still **not** an injection site: prepending to it would produce a shape no adapter reads, so such a message is left for the next candidate. `array_is_list`, not `is_array`.

**The silence.** A composed block that finds no user message is now logged. The list is still returned unchanged — the block is never escalated into the system role to satisfy a missing user turn — but skills carry instructions and constraints, and "silently absent" is the failure mode that matters. A call with **no** skills stays silent, so the warning marks a lost block rather than every skill-less call.

## Scope, since the issue asks

The issue's "not yet established" point stands, and the answer is not reassuring in either direction:

**No in-repo caller produces a multimodal message on the skill path.** The two callers of `augmentMessages()` are `LlmServiceManager::chat()` and `ToolLoopService`. The only place that builds image parts is `VisionService`, which dispatches through `llmManager->vision()` with `VisionContent` objects and never reaches skill injection — exactly as the issue said.

So the reachability is the **public API's**: `chat()` and `chatForConfiguration()` accept `array<string, mixed>` messages, and `MessageShaper` deliberately passes the multimodal shape through. This fixes the API surface, not an observed call. Stated plainly rather than dressed up as a live incident.

## Related Issue

Closes #645.

## Type of Change

## Checklist

- [ ] ADR — none: this makes the injection do what its own docblock describes

### Tests

Written before the fix, and both were red against unchanged `Classes/`:

- a multimodal first turn receives the block as a leading text part, the original parts survive in order, and the later text-only turn stays untouched — one injection site, not two;
- a composed block with no user message warns.

A third test is green both ways on purpose: **no** skills means no block and no warning. Without it, a change that warned on every call would satisfy the second test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,31 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- The skill block reaches a multimodal user turn (#645).
+  `SkillInjectionService` recognised an array-shaped message as a user
+  message only when its `content` was a **string**, so a conversation whose
+  first user turn carries content as a list of parts had the block prepended
+  to some later text turn — or, with no later text turn, not at all. Silently.
+  Multimodal content is a supported input shape, not an accident:
+  `MessageShaper::normalise()` converts only the exact 2-key string/string
+  form into a `ChatMessage` and passes everything else through for the
+  adapters to convert. The block is now prepended as a leading
+  `{type: 'text'}` part, the shape every adapter reads
+  (`ClaudeProvider::convertMultimodalContent()` translates it into Claude's
+  own format). Existing parts keep their order and are never merged into,
+  because a part may be an image.
+  An associative content array is still not an injection site — prepending
+  to it would produce a shape no adapter reads — so such a message is left
+  for the next candidate.
+- A composed skill block that finds no user message is logged instead of
+  vanishing. The message list is still returned unchanged, since the block is
+  never escalated into the system role, but skills carry instructions and
+  constraints and "silently absent" is the failure mode that matters. A call
+  with no skills stays silent — the warning marks a lost block, not every
+  skill-less call.
+
 ## [0.26.0] - 2026-08-06
 
 ### Added
--- a/Classes/Service/Skill/SkillInjectionService.php
+++ b/Classes/Service/Skill/SkillInjectionService.php
@@ -103,6 +103,20 @@ final readonly class SkillInjectionService
             $augmented[] = $message;
         }
 
+        if (!$injected) {
+            // The list is still returned unchanged — the block is never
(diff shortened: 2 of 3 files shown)


CybotTM (commented):

### Randfälle nachgerechnet

- **`array_is_list` statt `is_array`.** Ein assoziatives `content`-Array ist kein Part-Array; ein vorangestellter Text-Part erzeugte dort eine Form, die kein Adapter liest. Solche Nachrichten bleiben Kandidat für die nächste Runde statt zur Injektionsstelle zu werden. Richtige Richtung.
- **Liste aus Nicht-Arrays** (etwa `['a','b']`) würde `array_is_list` passieren und einen Text-Part vorangestellt bekommen. Der Adapter macht daraus nichts Sinnvolles — allerdings genauso wenig aus der Liste selbst. Der Fall ist schon vor der Injektion kaputt, und ihn hier abzufangen hieße, eine Form zu validieren, für die `MessageShaper` bewusst nicht zuständig ist.
- **Leere Part-Liste** → nur der Block. Zulässig.
- **Die Warnung feuert nicht ohne Block.** Der dritte Test hält genau das fest, und er ist in beiden Zuständen grün — ohne ihn hätte eine Änderung, die bei *jedem* Aufruf warnt, den zweiten Test erfüllt.

### Zur Reichweite

Der PR-Text sagt ausdrücklich, dass **kein Aufrufer im Repo** diese Form auf dem Skill-Pfad erzeugt und die Erreichbarkeit an der öffentlichen API hängt. Das habe ich nachgeprüft: `VisionService` läuft über `llmManager->vision()` mit `VisionContent`-Objekten und erreicht die Injektion nie. Der Fix repariert die API-Oberfläche, keinen beobachteten Vorfall — so steht es dort auch, und so sollte es stehen bleiben.

Kein Fund.

PR #743, opened 2026-08-11
feat(mcp): a conformance pack every supported MCP connection is held to

CybotTM:
The plan's §9, held to its own narrow definition of done — *not* "nr_llm can do everything MCP can", but **everything nr_llm supports as an MCP client is policy-, audit- and health-integrated**. **Stacked on #736**; review that first.

One abstract case carries 34 checks; a concrete case supplies only an `McpConnectionProfile` — what legitimately differs between two servers. Every response, failure and bound is scripted identically through a faked PSR-18 client; **no live network anywhere**. A new transport is a profile plus a three-line subclass. Every item on the plan's checklist is classified in ADR-161 as covered / covered elsewhere / out of scope, with the reason.

**Two real defects, found by writing the checks**

- A transport failure *and* MCP's own tool-level `isError` were both persisted as **successful** steps whose content merely read like an error. "How often does this server fail" was a string search over tool output. `McpClient::callTool()` now returns a typed outcome and both become `ToolResult::error()`.
- Non-text content blocks were dropped in silence. The answer now **opens** with a note counting them and naming their types — it leads rather than trails because `ToolResultBounder` truncates at 50 000 bytes, so the answers most likely to be cut are exactly the ones where the note matters.

**Cancellation is a gap and is written down as one.** It does not exist: `AgentRuntime::cancel()` flips persisted state and the transport has no abort path, so a cancelled run still blocks up to 15 s per outstanding leg. The check that guards it asserts an **absence** — no cancellation seam on the client or the transport — and was probed to go red the day one is built. The first version was a fake client with a closure that raised nothing and could not have failed; three documents claimed it pinned behaviour. It does not ship that way.

**Audit keeps no second stream, and that is a decision.** Every imported MCP tool declares a write effect, so ADR-141's fence refuses it in a segment holding no persisted run and no lease *before* the call happens — "an MCP tool executed and nothing was recorded" is not reachable through the runtime. A parallel client-side log would answer the same question twice and disagree the first time one failed. `tx_nrllm_governance_event` stays denial-only: an executed call is not a policy event.

**Scope boundary held:** nr_llm remains an MCP *client*. Nothing here exposes TYPO3 as an MCP server.

Gates: `cgl`, `rector` (applied, both dry-runs stable), `phpstan`, `unit` on both a PHP 8.2 resolve (PHPUnit 11) **and** a PHP 8.3 resolve (PHPUnit 12.5, the version six CI cells use), CHANGELOG check, plus the MCP functional classes — all green.

ADR-161.


CybotTM:
## Review note — Copilot unavailable, reviewed by the author

The `copilot_code_review` ruleset is active and Copilot cannot deliver: the requesting account is out of quota for this month, and a quota ceiling does not clear by asking again. `pr-status.sh` keeps reporting `request-review` by design — it cannot observe that the diff was read.

So this PR was reviewed against its final head instead, adversarially and independently of the round that preceded the first push. It returned **1 major, 1 minor**.

What it found:

- **[major]** `Documentation/Adr/Adr161McpClientConformance.rst` — Two rows of the ADR's checklist — "data classification" and the second half of "invalid schema" — are listed as Covered, but the behaviour they describe lives in `McpToolProvider::tools()`, which no test in the repository exercises. The suite asserts the McpTool constructor arguments back to itself.
- **[minor]** `Classes/Service/Tool/Mcp/McpClient.php` — The omission note prepended to every tool result carries up to 160 characters of server-chosen text inside a first-person `[nr_llm …]` frame, so "the note is ours, not the server's" is only true of the sentence around it — and none of the four bounds the CHANGELOG names (charset strip, 32-char clip, five-type cap, `unknown` fallback) has a test.

**2 fixed** in the commit above.

Stated plainly because it matters for how much this is worth: this is the author reviewing their own change. It is not a substitute for a second pair of eyes, and it is what was available.

PR #745, opened 2026-08-11
feat(setup): a use-case pack, and an onboarding step that starts from the use case

CybotTM:
The plan's §14 — its lowest-scored area (6/10) and the one it argues holds more differentiation than the kernel.

**Setup no longer has to start at "Provider".** An earlier step asks what the operator wants to do and suggests a configuration, a governance posture, actions, skills and snippets. The existing technical wizard is untouched behind it.

**One pack ships end to end** — Editorial Starter — rather than a framework with nothing in it: configuration preset, task definitions, skills, snippets, a *recommended* governance posture, required capabilities, optional actions. Recommended, not applied: ADR-145 argued down an apply path and this does not reopen it.

**Nothing is written before an operator confirms**, and the plan screen names every effect they are confirming — including the two that are not new records:

- The **snippet-tag link** is written on either creation path and only ever *adds* to the operator's own selection. The review found that a configuration imported through the preset UI first — a path this diff itself creates — got the pack's snippets and never the link: composed into nothing, no error, and the pack read as fully installed so the confirm button disappeared. Fixed, and proved with a test that walks that exact sequence and asserts the resolver actually returns both snippets.
- A tag selection is **not scoped to one configuration**, so installing can change the prompts of pre-existing configurations that already select the same tags. The screen lists them by name before the write, because an operator cannot confirm an effect they are not shown.

**Re-installing is safe:** "already installed" is per record, nothing is overwritten, and the tag union is the single documented exception. A union that would overflow the column refuses the install *before* any task or snippet is created, rather than truncating the operator's tags.

**Not built, and said so:** the Translation, Media, Local-AI and Enterprise-Cloud packs. One complete pack plus the entry step is the honest scope; a registry with one entry is still a registry, and the next pack is data.

Gates: `cgl`, `rector` (applied, both dry-runs stable), `phpstan`, `unit` on a PHP 8.2 resolve and a PHP 8.3 resolve (PHPUnit 12.5), CHANGELOG check, plus both new functional classes — all green.

ADR-163.


CybotTM:
## Review note — Copilot unavailable, reviewed by the author

The `copilot_code_review` ruleset is active and Copilot cannot deliver: the requesting account is out of quota for this month, and a quota ceiling does not clear by asking again. `pr-status.sh` keeps reporting `request-review` by design — it cannot observe that the diff was read.

So this PR was reviewed against its final head instead, adversarially and independently of the round that preceded the first push. It returned **2 major, 4 minor**.

What it found:

- **[major]** `Classes/Service/UseCase/UseCasePackInstaller.php` — The plan screen shows the tags an install ADDS to the pack's configuration, but never the pre-existing operator snippets those tags pull INTO that configuration — which can raise its injected-context data class and make a previously-working configuration fail closed at send time. The PR body's "the plan screen names every effect they are confirming" and ADR-163's symmetric warning for the other direction do not hold for this direction.
- **[major]** `Classes/Service/UseCase/EditorialStarterPackProvider.php` — The pack's configuration preset is described to the operator as having "no tool access", but an empty `allowedToolGroups` means the opposite in this codebase: no group restriction at all.
- **[minor]** `Classes/Controller/Backend/UseCasePackController.php` — The success flash reports the pack's configuration as "already present and left unchanged" in exactly the runs where the installer wrote its `snippet_tags`.
- **[minor]** `Documentation/Administration/Index.rst` — The new admin module "Get Started" is absent from the documentation's module inventory and toctree, unlike every other `nrllm_*` submodule.
- **[minor]** `Resources/Private/Templates/Backend/Configuration/List.html` — `id="nrllm-configuration-presets"` is added with no consumer anywhere in the diff or the repository.
- **[minor]** `Resources/Private/Templates/Backend/Task/Execute.html` — The contrast fix for `text-bg-danger` is applied to two badges and skipped on four others, including the byte-identical "No model assigned" badge on another page — the template comment states the rule as general.

**6 fixed** in the commit above.

Stated plainly because it matters for how much this is worth: this is the author reviewing their own change. It is not a substitute for a second pair of eyes, and it is what was available.

PR #795, opened 2026-08-16
docs: point the four dead path references at the files that exist

CybotTM:
## Description

`AGENTS.md` named three paths that do not resolve, and the landing page's security-controls data named a fourth. An agent that follows them reads nothing, and in the two ADR cases it silently skips the rationale it was sent to find — the failure mode that leaves no trace, because nothing errors when a documentation pointer is dead.

Each replacement was verified by reading the target, not by fixing the filename to the nearest match. Two of the four were wrong in a way a spelling correction would have got wrong as well:

| Reference in | Named | Actually | Why |
|---|---|---|---|
| `AGENTS.md:119` | `Adr/Adr001ThreeTierProviderArchitecture.rst` | `Adr/Adr013ThreeLevelConfigurationArchitecture.rst` | ADR-001 is *Provider Abstraction Layer*. The record that decides Provider → Model → Configuration is ADR-013, so the number was wrong, not the spelling. |
| `AGENTS.md:119` | `Classes/Provider/AGENTS.md` | `Classes/AGENTS.md` | No `AGENTS.md` exists below `Classes/`. `Provider/Contract/` is covered by `Classes/AGENTS.md`. |
| `AGENTS.md:161` | `Adr/Adr012ApiKeyStorageVault.rst` | `Adr/Adr012ApiKeyEncryption.rst` | The file exists under a different name **and** its status is `Superseded`. [`Adr/Index.rst`](https://github.com/netresearch/t3x-nr-llm/blob/main/Documentation/Adr/Index.rst) states that the nr-vault integration replacing it "no record decided", so the bullet now says that rather than implying a live ADR stands behind the rule. |
| `landingpage/build/data/project.json:68` | same dead ADR filename, as the `evidence` URL of the `vault-keys` security control | `Adr012ApiKeyEncryption.rst` | This one renders as a public 404 on the landing page. Every other `evidence` path in that file resolves — checked all nine. |

## Related Issue

## Type of Change

- [x] Documentation update

## Checklist

- [ ] I have run `make gate` — not run, and it would not cover this: the diff is two Markdown lines and one JSON string, touching no path PHPStan, Rector or any PHPUnit suite analyses. `composer ci:test:repo` (badges, changelog, workflow ownership) passes.
- [ ] I have added tests that prove my fix/feature works — no behaviour changed. The four references were each verified against the file on disk.
- [ ] I have added a `CHANGELOG.md` entry — docs-only commits in this repo do not carry one (checked the last eight `docs:` commits).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,7 +116,7 @@ The full `./Build/Scripts/runTests.sh -s functional` run includes ~34 provider-c
 <!-- AGENTS-GENERATED:START directory-structure -->
 ## Architecture
 
-Three-tier model: **Provider → Model → Configuration**. See `Documentation/Adr/Adr001ThreeTierProviderArchitecture.rst` for the design rationale and `Classes/Provider/AGENTS.md` for adapter contracts.
+Three-tier model: **Provider → Model → Configuration**. See `Documentation/Adr/Adr013ThreeLevelConfigurationArchitecture.rst` for the design rationale and `Classes/AGENTS.md` for adapter contracts.
 
 ### Directory Structure
 ```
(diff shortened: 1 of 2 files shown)


CybotTM:
Copilot was unavailable for this PR — the review quota is exhausted for the month, so re-requesting cannot change it before the reset. Reviewed by hand instead, and recording what was actually checked rather than a verdict.

Each of the four replacements was verified by opening the target, not by matching the nearest filename. That mattered twice: ADR-001 is *Provider Abstraction Layer*, so the three-tier sentence needed ADR-013 rather than a corrected spelling of 001; and `Adr012ApiKeyEncryption.rst` carries `:Status: Superseded`, which is why the bullet now says so instead of pointing at it as if it were live. `Documentation/Adr/Index.rst` line 48 is the source for "no record decided" — quoted rather than inferred.

Both AGENTS.md edits sit inside `AGENTS-GENERATED` blocks whose header says to edit content and not structure; both change content only. Checked that no generator exists in the repository that would regenerate those blocks and revert this — there is no reference to `AGENTS-GENERATED` in `Build/` or `.github/`.

Swept for the old names afterwards across the whole tree: no occurrence remains. All nine `evidence` paths in `project.json` now resolve; before this change exactly one did not.

PR #796, opened 2026-08-16
chore: check every AGENTS.md scope carries its agent aliases

CybotTM:
## Description

Seven of the eight directories holding an `AGENTS.md` carry `CLAUDE.md` and `GEMINI.md` beside it as symlinks to that file. `Documentation/` carried neither. This PR started as "close that gap"; CI showed the gap is load-bearing, and what it now adds is the check plus the exemption that explains it.

**What the first attempt found.** Adding the two symlinks broke all three `Render Documentation` jobs. The TYPO3 docs renderer walks that tree through Flysystem, whose local adapter throws `SymbolicLinkEncountered` on any symlink it lists — it does not degrade the render, it aborts it. `Documentation/` was almost certainly the one directory without the pair for that reason, so calling its absence a defect was too quick. Excluding the path is not this repository's to do: the render comes from a shared `netresearch/*` reusable workflow with no input for it, serving every caller.

**What ships.** `Build/Scripts/check-agent-doc-symlinks.php` requires, for every directory holding an `AGENTS.md`, both aliases present as symlinks whose target is `AGENTS.md`. It rejects a missing alias, a regular file standing in for one, and a symlink pointing elsewhere. `Documentation/` is a named exemption carrying its cause in the source, in the shape `check-workflow-ownership.php` already uses for the aggregate gate, and the failure message points at `EXEMPT` so nobody drops an alias silently instead.

**Why symlinks and not copies.** A copy is a second source that drifts in silence: the two files disagree, nothing notices, and which one an agent believes depends on its vendor. The symlink makes the question unaskable — which is why the check enforces the shape rather than mere presence.

**Does the original gap stay open?** Partly, and knowingly. The root `AGENTS.md` does carry the aliases and names `Documentation/AGENTS.md` in its index table, so a reader who opened `CLAUDE.md` is told the file exists and where. That is weaker than an alias and it is what the renderer permits.

**Probed, not assumed.** Each of the three defect shapes was injected against the working tree and the check failed on each, naming the file and the reason; the probes were reverted and the revert verified. After the exemption was added, removing `Tests/CLAUDE.md` still fails the check — so the exemption narrowed the rule without disarming it.

## Related Issue

## Type of Change

- [x] Refactoring (no functional changes)

## Checklist

- [ ] I have run `make gate` — the part that applies is green: `cgl` passes (it covers `Build/`), and all four `ci:test:repo` scripts pass. PHPStan and Rector do **not** analyse `Build/Scripts` — their configured paths are `Classes`, `Configuration`, `Tests`, `Resources` and `ext_*.php` — and no PHPUnit suite covers it, so the remaining gate steps have nothing to say about this diff. Saying that rather than reporting a gate I did not run.
- [x] I have added tests that prove my fix/feature works — the check is the test; its three failure modes were provoked and observed, and re-probed after the exemption.
- [x] I have updated the documentation accordingly — the script header carries the rationale and the measured renderer error.
- [ ] I have added a `CHANGELOG.md` entry — no user-facing change; this is repo tooling.

### DIFF
--- a/Build/Scripts/check-agent-doc-symlinks.php
+++ b/Build/Scripts/check-agent-doc-symlinks.php
@@ -1,0 +1,107 @@
+#!/usr/bin/env php
+<?php
+
+/*
+ * Copyright (c) 2025-2026 Netresearch DTT GmbH
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+declare(strict_types=1);
+
+/*
+ * Refuse an AGENTS.md whose CLAUDE.md / GEMINI.md siblings are missing.
+ *
+ * Agents read whichever filename their vendor taught them, so every scoped
+ * AGENTS.md needs the two aliases beside it or that scope's rules are invisible
+ * to whoever asks for the other name. The aliases are symlinks to AGENTS.md and
+ * not copies on purpose: a copy is a second source that drifts silently, and
+ * nothing here would notice — the two files simply disagree, and which one an
+ * agent believes depends on its vendor.
+ *
+ * The gap this check closes was real rather than hypothetical.
+ * `Documentation/` was the one directory of eight without the pair, so an agent
+ * reading CLAUDE.md got the RST conventions, the ADR format and the
+ * `Adr/Index.rst` lifecycle from nowhere, while an agent reading AGENTS.md got
+ * all three. Nothing failed; the rules were just absent for half the readers.
+ *
+ * Runs in the shared workflow's `repo-checks` job via `ci:test:repo`, which is
+ * also what pre-commit runs, so the local and CI halves are the same command.
+ */
+
+const ALIASES = ['CLAUDE.md', 'GEMINI.md'];
+
+/** Directories that never hold source we govern. */
+const SKIP = ['.git', '.Build', 'vendor', 'node_modules', 'var', 'public'];
+
+$root = dirname(__DIR__, 2);
+
+/**
+ * @return list<string> every directory below $root holding an AGENTS.md
+ */
+function agentsDirectories(string $root): array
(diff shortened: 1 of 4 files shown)


CybotTM:
Copilot was unavailable for this PR — the review quota is exhausted for the month, so re-requesting cannot change it before the reset. Reviewed by hand instead; what follows is what was checked, including the limits I did not close.

The check was probed rather than trusted: missing alias, regular file in place of a symlink, and symlink pointing elsewhere each produce a failure naming the file and the reason, and after `Documentation` became exempt, removing `Tests/CLAUDE.md` still fails — so the exemption narrowed the rule without disarming it. Probes were reverted and the revert verified against `git status`.

The exemption is the substance of this PR and it is an admission: adding the two symlinks aborted all three `Render Documentation` jobs with `League\Flysystem\SymbolicLinkEncountered`. That error is quoted verbatim in the script header so the next person does not rediscover it by breaking CI. Excluding the path would mean changing a shared reusable workflow that serves every caller, which does not belong in this repository.

Three limits worth naming rather than leaving implicit. The check only inspects directories that already hold an `AGENTS.md`, so a stray `CLAUDE.md` elsewhere is not caught. A symlink whose target does not exist would pass, though it cannot occur here because the directory qualifies precisely by holding the target. And `Documentation/` keeps the gap the PR set out to close — the root `AGENTS.md` naming it in the index table is weaker than an alias, and it is what the renderer permits.

Gate coverage: `cgl` passes and covers `Build/`; all four `ci:test:repo` scripts pass. PHPStan and Rector do not analyse `Build/Scripts`, and no PHPUnit suite covers it, so nothing else in `make gate` has anything to say about this diff.

PR #797, opened 2026-08-16
docs: point the PR checklist at the gate the workflow documents

CybotTM:
## Description

The checklist asked contributors to confirm `composer ci`, which is not the pre-push gate this project documents. The two sets differ in exactly the place that matters:

| | cgl | phpstan | unit | integration | fuzzy | rector | functional | changelog |
|---|---|---|---|---|---|---|---|---|
| `composer ci` | ✓ | ✓ | ✓ | ✓ | ✓ | — | — | — |
| `make gate` | ✓ | ✓ | ✓ | — | ✓ | ✓ (pinned 8.2) | ✓ (sqlite) | ✓ |

Rector and the functional suite are precisely the two the `Makefile` comment names as the ones that get skipped when the suites are invoked by hand, and which the CI matrix then finds a push later. So the box a contributor ticked certified a narrower set than [`AGENTS.md`](https://github.com/netresearch/t3x-nr-llm/blob/main/AGENTS.md) prescribes, and the gap was invisible at the moment of ticking it.

Both targets stay — `AGENTS.md` says they answer different questions — but the checklist now names the one that gates a push.

## Related Issue

## Type of Change

- [x] Documentation update

## Checklist

- [ ] I have run `make gate` — not run, and it does not apply: the diff is one line of Markdown in a PR template, touching no path any suite analyses. `composer ci:test:repo` passes.
- [ ] I have added tests that prove my fix/feature works — no behaviour changed; the claim in the new line was verified against the `gate` and `ci` targets in the `Makefile` and the `ci` script in `composer.json`.
- [ ] I have added a `CHANGELOG.md` entry — docs-only, and this repo's `docs:` commits do not carry one.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,7 +17,7 @@
 ## Checklist
 
 - [ ] My code follows the project's coding standards
-- [ ] I have run `composer ci` and all checks pass
+- [ ] I have run `make gate` and all checks pass (not `composer ci` — that set omits Rector and the functional suite)
 - [ ] I have added tests that prove my fix/feature works
 - [ ] I have updated the documentation accordingly
 - [ ] I have added a `CHANGELOG.md` entry under `## [Unreleased]`


CybotTM:
Copilot was unavailable for this PR — the review quota is exhausted for the month, so re-requesting cannot change it before the reset. Reviewed by hand instead.

The claim added to the checklist line was verified against both definitions rather than taken from prose. `composer ci` resolves to cgl, phpstan, unit, integration and fuzzy. The `gate` target is `changelog cgl phpstan test-unit test-fuzzy` plus rector pinned to PHP 8.2 and functional against sqlite. So "omits Rector and the functional suite" is accurate, and it is deliberately not exhaustive — `ci` also omits the CHANGELOG check and additionally carries `integration`, which the gate does not. The line names the two the Makefile comment itself calls out as the ones that get skipped by hand.

Nothing else changes: both targets remain, and this only redirects which one the checklist certifies.

Issue #770, opened 2026-08-13
No per-call cost or real token counts: telemetry and cost cannot be joined


CybotTM:
`tx_nrllm_telemetry` records routing, success, latency and a complexity estimate per provider call, but carries no cost and no real token counts. `routing_signal_cost` is a ranking signal, not money; `complexity_tokens` is explicitly an estimate. Actual cost lives in `tx_nrllm_service_usage` as a **daily aggregate** with no `correlation_id`, so there is no join key from a cost to a call, a request shape or a complexity bucket.

Consequence: any per-bucket or per-arm cost statement — "median cost per complexity bucket", "real cost drops" as ADR-156's third activation criterion — is not computable. Not hard to compute: not computable.

A second defect sits underneath it. `$usage->estimatedCost` is never set by any adapter (`AbstractProvider.php:651-663` is dead on that field); cost is derived solely from the model row's pricing (`UsageMiddleware.php:185-198`). Models priced at zero — Ollama, Groq, local — therefore report a cost of 0 by construction, which is precisely the arm any cheap-model experiment is about.

**What closing it needs**

Per provider call, correlated by the id the call already carries:

- actual input tokens, where the provider reports them;
- actual output tokens, where the provider reports them;
- the model actually served;
- the cost derived from those;
- retry count;
- `correlation_id`.

Where a provider reports no usage the value is `NULL`, never `0` — a missing measurement and a measured zero must stay distinguishable, which is exactly what the current zero-priced case gets wrong.

Precondition for ADR-156's third activation criterion, and for any routing readiness report that mentions cost.

Ref: ADR-156.

PR #798, opened 2026-08-16
chore(spec-kit): proof of concept — specify, plan and tasks against #770

CybotTM:
## Description

Proof of concept, deliberately a draft: an experiment to judge, not a change to merge. Spec Kit v0.16.4 with the `lean` preset and the `claude` integration, run through `specify → plan → tasks` against a real issue, [#770](https://github.com/netresearch/t3x-nr-llm/issues/770). No implementation — the point is what the three stages return before code exists.

The question it answers: does specifying a cross-layer change first produce fewer architecture corrections and fewer scope misunderstandings than going from issue straight to code?

## What the three stages returned

**The specification found that #770 mislocates its own second defect.** The issue places "zero-priced models report a cost of 0 by construction" at `UsageMiddleware.php:185-198`. Those lines behave correctly: no adapter sets `estimatedCost`, `Model::hasPricing()` is `costInput > 0 || costOutput > 0` and so returns false for a zero-priced model, the derivation is skipped, and the middleware omits the `cost` key rather than writing a zero. The zero enters at persistence, via `$metrics['cost'] ?? 0.0` in `UsageTrackerService` lines 131 and 154. The conclusion in #770 stands; its location does not, and a plan written against the named lines would have changed correct code and left the cause in place.

**The plan found that three of the six requirements are already partly satisfied.** `tx_nrllm_telemetry` carries `correlation_id` with an index and a UUID minted per call, plus `served_provider` and `served_model` naming the configuration that answered. Planning from the issue text alone would have added columns that exist.

**It also found where the record belongs, from a convention rather than a preference.** That same table already keeps `complexity_tokens` as `DEFAULT NULL`, commented "NULL where no context fit ran, which is not the same as an empty send" — the null-is-not-zero distinction the spec demands is this table's existing habit. `tx_nrllm_service_usage` is the opposite: `estimated_cost` is `NOT NULL DEFAULT '0.000000'` with a width deliberately aligned to the budget-ceiling columns, so a nullable per-call cost cannot live there without changing budget arithmetic. The plan therefore accepts in writing that the daily aggregate keeps writing `0.0` for zero-priced models, rather than discovering that trade-off during implementation.

**And it answered the public-surface question before implementation, which is why the spec deferred it there.** `TelemetrySignals` is `@api` and frozen in `Tests/Unit/Api/api-surface.txt` at line 1064 with every method and property listed, so a recorder method plus properties is additive — regenerate the snapshot, add `### Added` — not a break. Discovering that at the point `api-surface.txt` fails is the normal way, and it is one round of rework later.

**One question is left open on purpose and blocks the task list.** FR-003 does not say whether a retry is a fallback hop or a provider-level retry. `fallback_attempts` counts hops. If FR-003 means hops it is already satisfied and T012 disappears; if it means retries, the count exists nowhere. Answering it silently would produce a column whose meaning nobody could state.

## Deliberate boundaries

| Decision | Reason |
|---|---|
| No `git` extension | Branching, worktrees, signed commits and the pre-push gate live in `AGENTS.md`. Verified: `create-new-feature.sh` makes no git call of its own. |
| Constitution kept to seven principles | Its first says `AGENTS.md` outranks it. Six of seven constrain what a *specification* must satisfy, not how code is written. |
| `/.specify` and `/specs` in `export-ignore` | Neither was excluded, so every specification would otherwise ship in the composer and TER package. Verified by `git archive`: both absent, while `Classes/` contributes 814 entries and `Documentation/` 279 — the probe can tell the difference. |
| `.claude/skills` untracked | `.gitignore:12` already ignores `.claude/`, and the ten skills regenerate deterministically from the tracked `.specify/presets/lean`, differing only by generated frontmatter. |

## Two findings about the tool itself

**The bundled script silently opts out of the preset.** `create-new-feature.sh` produced the full 131-line template with placeholders even with `lean` installed, because it resolves `.specify/templates/spec-template.md` while `lean` replaces the *commands*, which write the artefact directly from a 23-line prompt. Nothing indicates which route you are on, and classic Spec Kit habits lead straight to the ceremony `lean` exists to remove.

**The bash scripts need `python3` with PyYAML on `PATH`.** They call bare `python3`; where that resolves to a venv without PyYAML the run stops at `Error: PyYAML is required to resolve preset template composition`.

## Reproducing

```
uv tool install specify-cli --from git+https://github.com/github/spec-kit.git@v0.16.4
specify init --here --force --integration claude --preset lean
```

## Honest limit

Three findings before any code is not yet the comparison this PoC was set up for — that needs the counterfactual, which is what a normal issue-to-code run would have cost. Two of the three (the mislocated defect, the already-present columns) would plausibly have surfaced in review or in implementation instead, one round later each. That is the case for the pipeline, not proof of it.

## Type of Change

- [x] Documentation update

## Checklist

- [ ] I have run `make gate` — not applicable: this diff adds `.specify/`, `specs/` and two `.gitattributes` lines, touching no path any suite analyses. `composer ci:test:repo` passes.
- [ ] I have added tests that prove my fix/feature works — no behaviour changed. The two `export-ignore` entries were verified against `git archive`; every code claim in the spec and plan was verified by reading the cited file.
- [ ] I have added a `CHANGELOG.md` entry — no user-facing change, and the PoC may be reverted.
- [ ] I have added an ADR — none yet. If Spec Kit stays, adopting it is exactly the kind of decision that needs one.


CybotTM:
Copilot was unavailable for this PR — the review quota is exhausted for the month, so re-requesting cannot change it before the reset. Reviewed by hand instead.

**Blast radius.** Nothing executable changes. The diff adds `.specify/`, `specs/`, two `.gitattributes` lines and the three specification artefacts; it touches no path that PHPStan, Rector or any PHPUnit suite analyses, and no runtime code. `composer ci:test:repo` passes. The one change with a functional consequence is the `export-ignore` pair, and that was verified against `git archive` rather than assumed: `.specify/` and `specs/` contribute zero entries while `Classes/` contributes 814 and `Documentation/` 279 — so the probe can distinguish, and the specifications stay out of the composer and TER package.

**Every code claim in the artefacts was read at the source, not carried over from the issue.** That is what produced the substance: #770 locates its second defect at `UsageMiddleware.php:185-198`, and those lines behave correctly — the zero enters at `UsageTrackerService.php:131` and `:154` via `$metrics['cost'] ?? 0.0`. Likewise `hasPricing()` is `costInput > 0 || costOutput > 0`, which is why the middleware omits the key rather than writing a zero, and `TelemetrySignals` is `@api` and frozen in `api-surface.txt` at line 1064, which makes the proposed change additive rather than breaking.

**Where the artefacts commit us to something, it is stated as a decision with its consequence.** The per-call record goes into `tx_nrllm_telemetry` and not the daily aggregate, and the accepted price — the aggregate keeps writing `0.0` for zero-priced models — is written down rather than left to be discovered during implementation.

**What this PR does not establish.** It does not win the comparison it was set up for. Three findings before any code is a case for the pipeline, not proof of it: two of the three would plausibly have surfaced in review or implementation instead, one round later each, and no counterfactual run was measured. The rule that came out of the exercise landed separately in #799 precisely so it does not depend on this trial surviving.

**Two findings about the tool, recorded because judging it was the point.** `create-new-feature.sh` produces the full 131-line template even with `lean` installed, because it resolves `.specify/templates/` while `lean` replaces the commands — so reaching for the script silently opts out of the preset. And the bundled bash scripts need a `python3` with PyYAML on `PATH`.

**One limitation of the merge itself.** `.claude/skills` stays untracked because `.gitignore:12` ignores `.claude/`. The pipeline is reproducible from this commit — the skills regenerate from the tracked `.specify/presets/lean`, differing only by generated frontmatter — but anyone cloning has to run the init command once. That is documented in the PR body rather than worked around.

Issue #793, opened 2026-08-14
ADRs cite code by line number and the citations rot silently — nothing checks them


CybotTM:
The two sibling issues (ADR-119's module count, ADR-108/094's builtin count) are instances. This is the mechanism behind them, and it has a second form that is worse because it *looks* precise: ADRs cite source locations as `File.php:NNN` or `File.php:NNN-MMM`, and a line number is invalidated by any edit above it in the file. There are 74 such citations across `Documentation/Adr/*.rst` and no check of any kind.

`Tests/Unit/AdrLifecycleTest.php` is the only automated guard on the ADR corpus. Its six tests cover status vocabulary, the `:Amended:`/`:Superseded:` pairing, and that `:ref:` targets resolve to a record that exists. It does not read the ADR body, so it sees none of this.

## Reproduce

```
grep -rhoE '\.php:[0-9]+(-[0-9]+)?' Documentation/Adr/*.rst | wc -l    # 74
```

## Confirmed rot, with the commit that caused it

`Documentation/Adr/Adr171PersonasTheCodeAlreadyAssumes.rst:115` cites `ResumeCoordinator.php:204` for the `mayActOnRun()` authorisation check.

Line 204 of `Classes/Service/Agent/ResumeCoordinator.php` is blank. The check is at line 205.

```
git show d3a8d718^:Classes/Service/Agent/ResumeCoordinator.php | sed -n '204p'
#         if (!$actor->mayActOnRun($run, ServiceAccountScope::AGENT_APPROVE)) {
sed -n '204,205p' Classes/Service/Agent/ResumeCoordinator.php
#
#         if (!$actor->mayActOnRun($run, ServiceAccountScope::AGENT_APPROVE)) {
```

The citation was correct when written. `d3a8d718` ("feat: four-eyes approval as a per-configuration switch") inserted the ADR-172 self-approval guard above it and shifted every line below by one. Nothing failed. The ADR still reads as a verified citation.

Off-by-one is the benign end. The same commit class moves a citation onto an unrelated statement, and a range citation (`:197-218`, `:375-381`) onto a different method entirely — at which point the record asserts something the cited code does not say, and the citation's precision is what makes a reviewer trust it.

## Why the counting issues are the same defect

[...]

PR #799, opened 2026-08-17
docs: write down the rule that every gate here presupposes

CybotTM:
## Description

Every gate in this repository runs on a diff that already exists. `make gate`, the api-surface snapshot, PHPStan, the CHANGELOG check and the ADR checkbox all need code before they can say anything. Measured: a grep over all eight `AGENTS.md` files finds no specification obligation — only false positives like "Specify PHP version" — and the documented Development Workflow begins with "Branch off `main`", which is after the design decisions have been taken. The repository has unusually strict post-implementation control and no pre-implementation control at all.

That gap is the actual defect. This PR writes the missing rule down, moves the one existing design obligation to where it belongs, and adds the part of it that can be mechanically enforced.

### 1. The rule (`AGENTS.md`, Development Workflow step 1)

Write down what the change must do, what it explicitly does **not** do, and which suite proves each requirement — before choosing where the code goes. With a table naming the classes of change it applies to, so it does not become "write a document for every typo":

| Change | Specify first |
|--------|---------------|
| Bugfix, dependency update, small internal refactor, documentation only | no |
| New provider | usually |
| New feature, new public API contract, breaking or deprecating change, security or credential topic, a change spanning several layers, large compatibility rework | yes |

**The format is deliberately not prescribed.** Which document that is remains open — [#798](https://github.com/netresearch/t3x-nr-llm/pull/798) trials one answer — and this rule should not depend on settling it. If that trial is dropped, the rule stands.

**And it says plainly that it is not machine-checked.** `ci:test:repo` sees the working tree, not the pull request; the job that could see one is `pr-quality`, which belongs to the shared `netresearch/.github` workflow and is not this repository's to extend. A rule that implies a gate it does not have is worse than one that admits it has none.

### 2. The ADR obligation moves before the implementation PR

It was a PR checkbox — "I have added an ADR … if this changes the public surface" — evaluated against a diff that already existed. The decision is what the implementation follows from, and `api-surface.txt` already tells you when the surface is being touched, so there is nothing to wait for. The PR template is updated to match, because a template contradicting `AGENTS.md` is exactly the defect [#797](https://github.com/netresearch/t3x-nr-llm/pull/797) just fixed.

### 3. `AdrReferenceIntegrityTest` — the enforceable part

`AdrLifecycleTest` already binds the records to each other: status words, the `:Amended:` / `:Superseded:` pairing, and that the index resolves every record it names. All of that stays inside `Documentation/Adr`. Nothing checked a reference from **outside** it — which is where four dead pointers sat until [#795](https://github.com/netresearch/t3x-nr-llm/pull/795), one of them published on the landing page as a security control's `evidence` URL, where it rendered as a public 404.

The test asserts existence only. It deliberately does not judge whether the record says what the citing sentence claims, because nothing mechanical can: both references #795 fixed pointed at the *wrong record*, not at a misspelling of the right one. It also does not address [#793](https://github.com/netresearch/t3x-nr-llm/issues/793) — a drifted line number still points at a line that exists.

**Probed in both directions.** Green on the current tree. With the pre-#795 `AGENTS.md` restored it fails, naming both dead references and nothing else. Re-proved after the Rector refactor below, because the earlier proof described code that no longer existed.

Two mistakes worth recording rather than quietly fixing. The first version failed on its own docblock: it scans `.php` and I had spelled the dead filenames out in full — the docblock now omits the extension and says why, so nobody "completes" them. And `ci / Rector` was red on the first push, because I carried over a conclusion from the sibling PR where the file sat in `Build/Scripts`, which Rector does not analyse; this one sits in `Tests/`, which `Build/rector/rector.php` lists explicitly. `LocallyCalledStaticMethodToNonStaticRector` wanted both private static helpers as instance methods, the second only after fixing the first cascaded into it.

## Related Issue

None. Found while reviewing this repository's governance.

## Type of Change

- [x] Documentation update

## Checklist

- [x] I have run `make gate` — in full, and it passes: `cgl`, `phpstan`, `unit`, `fuzzy`, `rector -n -p 8.2` and `functional -d sqlite`, plus the CHANGELOG check. It needed a dependency resolution at 8.2 in this worktree first: the `.Build` copied from `main` made the pinned Rector run die in `platform_check.php` before Rector started, and made the full unit suite fatal on `InMemoryVaultService` against a newer nr-vault than CI installs. Both disappeared after `runTests.sh -s composerUpdate -p 8.2`, which confirms by resolution what I had first only inferred — the fatal was the environment, not this branch.
- [x] I have added tests that prove my fix/feature works — the new test is the deliverable; it was verified red against the defect it exists for.
- [x] I have updated the documentation accordingly — `AGENTS.md` is the documentation here.
- [ ] I have added a `CHANGELOG.md` entry — no user-facing change; docs-only commits in this repo do not carry one.
- [ ] If this changes the public surface, the ADR landed before this PR — no public surface change.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -21,5 +21,5 @@
 - [ ] I have added tests that prove my fix/feature works
 - [ ] I have updated the documentation accordingly
 - [ ] I have added a `CHANGELOG.md` entry under `## [Unreleased]`
-- [ ] I have added an ADR under `Documentation/Adr/` if this changes the public surface
+- [ ] If this changes the public surface, the ADR under `Documentation/Adr/` landed before this PR (see step 1 of the Development Workflow in `AGENTS.md`)
 - [ ] My changes generate no new warnings
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -76,11 +76,34 @@ The full `./Build/Scripts/runTests.sh -s functional` run includes ~34 provider-c
 
 ## Development Workflow
 
-1. Branch off `main` (worktree convention — see project memory).
-2. Use `make` shortcuts (`make test-unit`, `make phpstan`, `make cgl`) — they delegate to `runTests.sh`.
-3. Pre-commit hooks via `Build/captainhook.json` (auto-installed by composer plugin) run cgl + phpstan + commit-msg checks.
-4. Sign commits with `git commit -S --signoff` (DCO required).
-5. **Pre-push gate: `make gate`.** It runs the six suites CI runs — `cgl`,
+1. **Specify before designing, for the classes of change below.** Write down what
+   the change must do, what it explicitly does **not** do, and which suite proves
+   each requirement — before choosing where the code goes. The format is not
+   prescribed here; writing it at all is the rule.
+
+   | Change | Specify first |
+   |--------|---------------|
+   | Bugfix, dependency update, small internal refactor, documentation only | no |
+   | New provider | usually |
+   | New feature, new public API contract, breaking or deprecating change, security or credential topic, a change spanning several layers, large compatibility rework | yes |
+
+   Why this step exists as a rule: every other gate in this repository runs on a
+   diff that already exists. `make gate`, the api-surface snapshot, PHPStan, the
+   CHANGELOG check and the ADR checkbox all presuppose code. Nothing before this
+   step checked anything, so a wrong premise about *what* was wanted survived
+   until review, and a wrong premise about the public surface survived until
+   `api-surface.txt` failed.
+
+   **This one is not machine-checked, and cannot be here.** `ci:test:repo` sees
+   the working tree, not the pull request; the job that could see one is
+   `pr-quality`, which belongs to the shared `netresearch/.github` workflow and
+   is not this repository's to extend. Keep it because it is the rule, not
+   because something will catch you.
+2. Branch off `main` (worktree convention — see project memory).
+3. Use `make` shortcuts (`make test-unit`, `make phpstan`, `make cgl`) — they delegate to `runTests.sh`.
+4. Pre-commit hooks via `Build/captainhook.json` (auto-installed by composer plugin) run cgl + phpstan + commit-msg checks.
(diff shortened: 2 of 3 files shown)


CybotTM:
Copilot was unavailable for this PR — the review quota is exhausted for the month, so re-requesting cannot change it before the reset. Reviewed by hand instead; below is what was actually checked and what I left open.

**The claim the PR rests on was measured, not asserted.** A grep for a specification obligation across all eight `AGENTS.md` files returns only false positives ("Specify PHP version", "Specify only `frontend`"), and the documented Development Workflow's first step is "Branch off `main`". The `pr-quality` job that could inspect a pull request resolves to `netresearch/.github`, which is why the rule says it cannot be gated from here rather than promising a check that does not exist.

**The gate ran in full, after a resolution at 8.2.** `cgl`, `phpstan`, `unit`, `fuzzy`, `rector -n -p 8.2`, `functional -d sqlite`, plus the CHANGELOG check. Worth recording that the first attempt could not run two of those: the `.Build` copied from `main` killed the pinned Rector run inside `platform_check.php` and made the unit suite fatal on `InMemoryVaultService`. Both vanished after `composerUpdate -p 8.2`, which settles by measurement what I had first only inferred from `main` failing identically.

**Limits of the new test, stated rather than implied.** It asserts that a referenced ADR file exists and nothing more — both references it was written for pointed at the *wrong record*, and no check can catch that. It scans a fixed extension list (`md rst json php yml yaml neon txt xml`), so a reference written in a format outside that list is missed. It skips symlinks, so the `CLAUDE.md` / `GEMINI.md` aliases are covered via their target rather than in their own right. And it does not address #793: a drifted line number still points at a line that exists.

**Two mistakes of mine are in the history rather than squashed out.** The first version failed on its own docblock, because it scans `.php` and I had written the dead filenames out in full. And `ci / Rector` was red on the first push, because I carried a conclusion over from the sibling PR — that file sat in `Build/Scripts`, which Rector does not analyse; this one sits in `Tests/`, which it does. I ran cgl and the tests and skipped Rector on a premise I had not re-checked.

The test was re-proved in both directions after the Rector refactor, since the earlier proof described code that no longer existed.

Issue #792, opened 2026-08-14
ADR-108 and ADR-094 count 40/41 builtin tools; there are 46


CybotTM:
Two records state the builtin-tool count as a load-bearing part of their reasoning, and both are now wrong by five. As with ADR-119's module count and ADR-130's approval-surface enumeration, the claim was true when written and nothing in the repo notices that it stopped being true.

The distinction that matters here: in ADR-094 the number is not colour, it is the *cost argument*. "Covers all 41 builtins without 41 edits" is the reason the contract change was deferred. A reader re-deriving that trade-off today is working from the wrong denominator.

## What the records say

`Documentation/Adr/Adr108TypedToolResultWithArtifacts.rst:85-86`

> The other 40 builtins ship text-parity via a mechanical ``ToolResult::text($string)`` wrap.

`Documentation/Adr/Adr108TypedToolResultWithArtifacts.rst:126`

> ``ToolInterface`` is a breaking change across all 41 builtins.

`Documentation/Adr/Adr094ToolDataClassTrustZone.rst:112-114`

> ``ToolInterface`` is **not** changed yet. Classifying by group plus seven explicit declarations covers all 41 builtins without 41 edits; promoting ``getDataClass()`` onto the contract is a later, announced breaking change, once observe-mode evidence exists.

## What the code says

`Classes/Service/Tool/Builtin/` holds 51 PHP files: 46 class declarations and 5 traits (`CollectsEnvironmentTrait`, `PlansOneEditorialWriteTrait`, `RendersTypoScriptTreeTrait`, `ResolvesLanguageLabelTrait`, `WritesThroughDataHandlerTrait`). None of the 46 is abstract.

Reproduce:

```
grep -hoE '^(final |readonly |abstract )*class [A-Za-z]+' Classes/Service/Tool/Builtin/*.php | wc -l   # 46
grep -lE '^abstract class' Classes/Service/Tool/Builtin/*.php | wc -l                                  # 0
grep -lE '^trait ' Classes/Service/Tool/Builtin/*.php | wc -l                                          # 5
```

So ADR-108's "the other 40" (41 total, one of them `ReadRecordsTool` as the sole v1 emitter) and both records' "all 41 builtins" undercount by five. `Classes/Service/Tool/Mcp/McpTool.php` also implements `ToolInterface` but is not a builtin — worth naming in whatever wording replaces the number, because it is the obvious thing a later reader miscounts in the other direction.

## Suggested fix

[...]

Issue #791, opened 2026-08-14
ADR-119 counts twelve submodules; there are fourteen, plus one outside the tree


CybotTM:
ADR-119 opens its Context by enumerating the module inventory, and the enumeration is out of date. The argument it carries — that the flat entries "read as a dumping ground" and that a section move would be expensive — is unaffected. The count is not decoration, though: the record is cited elsewhere *by that number*, and ADR-131's placement reasoning turns on the tree the number describes.

This is the same defect class as ADR-130 constraint 3, which is being amended in the branch `docs/adr130-amend`: a record states a fact about a list of registrations, the list grows, and nothing tells the record.

## What the record says

`Documentation/Adr/Adr119BackendModulePlacement.rst:18-20`

> nr_llm registers a parent module ``nrllm`` with twelve submodules — providers, models, configurations, tasks, snippets, skills, tools, playground, agent runs, analytics, setup wizard and overview — under TYPO3's **Administration** section.

Twelve names, and the same number is repeated at `:120` ("Twelve flat entries do not move as they are"), `:139` ("nr_llm's twelve submodules sit at the third") and `:141` ("or by grouping the twelve").

## What the code says

`Configuration/Backend/Modules.php` registers sixteen module keys. One is the parent (`nrllm`, `parent => 'tools'`). Fourteen are its children:

`nrllm_overview` (:77), `nrllm_providers` (:101), `nrllm_models` (:116), `nrllm_configurations` (:131), `nrllm_tasks` (:148), `nrllm_snippets` (:172), `nrllm_usecase` (:192), `nrllm_wizard` (:209), `nrllm_skills` (:224), `nrllm_tools` (:242), `nrllm_mcp` (:262), `nrllm_playground` (:280), `nrllm_runs` (:299), `nrllm_analytics` (:319).

The two the prose does not name are `nrllm_usecase` (use-case packs) and `nrllm_mcp` (MCP servers).

The sixteenth is `nrllm_aitasks` (`Configuration/Backend/Modules.php:342`), which is `parent => 'web'` — deliberately outside the `nrllm` tree, as its own comment and ADR-131 explain, because the menu hides a child whose parent's access check fails. So "twelve submodules" undercounts the tree by two and omits an nr_llm backend module entirely.

## Suggested fix

Rather than replacing twelve with fourteen and waiting for the next module to make it wrong again, state the inventory in a form that cannot silently drift:

[...]

PR #800, opened 2026-08-18
docs(adr): amend ADR-130's approval-surface constraint and gate it

### LINKED ISSUE #787: ADR-130 states a constraint that expired when the editor module shipped
ADR-130 constraint 3 states that `agent_approve` is **"doubly unreachable for non-admins today"** — both human approval surfaces sit behind admin gates — and that the `mayActOnRun()` branch "only becomes exercisable for non-admins with the editing module".

The editing module shipped. `Configuration/Backend/Modules.php:375-381` registers `AgentRunController::approve` and `submitInput` in `nrllm_aitasks`, which is `access => 'user'`. A non-admin holding the grant can reach the approval surface today.

**This is not a security defect.** A grant holder approving another user's suspended run is precisely what `AGENT_APPROVE` is for, and the write side is still authorised per run by `mayActOnRun()`. The behaviour is intended and the gate is doing its job.

**It is a documentation defect, and the repo names that class explicitly.** `Documentation/Adr/Index.rst` puts it plainly: *"An ADR that overturns part of an earlier one edits that earlier record's `:Status:` and `:Amended:` in the same change. An accepted ADR with an expired premise and a clean `Accepted` status is a defect, not history."* ADR-130 reads `Accepted` with no `:Amended:` field, while one of its stated constraints has expired. Anyone reasoning from it about what a non-admin can reach reaches the wrong conclusion.

The amendment belonged to whichever change made the surface reachable — ADR-131 by the look of it. That is worth noting for its own sake: the lifecycle rule depends on the amending author noticing, and here nobody did. If that has happened once it has plausibly happened elsewhere, and a sweep for accepted ADRs whose stated constraints no longer hold would be its own piece of work.

**What closing it needs**

`:Status:` and `:Amended:` on ADR-130 naming the record that made the surface reachable, and constraint 3 rewritten to describe what is actually true now: the grant is reachable, and the bound on it is `mayActOnRun()` plus the per-run authorisation rather than module access.

Found while deriving the personas the code assumes (ADR-171), where this appears as contradiction A.

CybotTM:
## Description

ADR-130 constraint 3 said `agent_approve` was "doubly unreachable for non-admins today". It named two surfaces, both behind admin gates — and **both still are**: `nrllm_runs` is `access => 'admin'` (`Modules.php:299-301`) and the Playground is too (`:280-282`) with `denyNonAdmin()` on its resume path.

**So the premise did not expire. A third surface arrived.** `nrllm_aitasks` is `access => 'user'` and registers `approve`/`submitInput` (`:342-344`, `:375-381`), so the enumeration stopped being exhaustive against a module that did not exist when it was written.

That distinction is the useful part. "The premise expired" reads as bad luck. **"A constraint that enumerates surfaces is only as durable as the enumeration"** is a failure mode a reviewer can guard against — and it is the one that bit. The record even predicted its own trigger ("becomes exercisable with the editing module") and the prediction left no mark anywhere a change had to pass.

Which record made it reachable was **established, not assumed**: `git log -S"nrllm_aitasks"` returns exactly one commit, `2ee16dfd feat: editor-facing AI Tasks module (ADR-131)`; `git log -S"AgentRunController::class"` returns that and the earlier admin-only module; and ADR-131's own Consequences already said it in prose — *"`agent_approve` through the shared approvals inbox (previously doubly unreachable for non-admins)"*. The author saw the consequence and wrote it down. Only the two lifecycle fields were missed. They are written now, from both ends.

## What the constraint says instead

**Not module access alone.** Both the module tick in `be_groups` and the grant are required, and neither substitutes for the other (ADR-131 decision 2).

Then three checks on the approve path, in the order `ResumeCoordinator` evaluates them — which is not the order I first wrote, and the code was allowed to decide:

1. `mayActOnRun()` (`:205`) — may this actor act on this run at all
2. ADR-172's `requiresSecondApprover` (`:222-226`) — optional, refuses an approval from the run's own initiator
3. ADR-133's `approverRefusal()` (`:314`) — withholds the release of a write the approver could not run themselves

Only the third releases the run, because only it had claimed anything. `submitInput`'s own gates are recorded **beside** the list rather than merged into it: it has no four-eyes check, and its `submitterRefusal` examines every pending call rather than only those declaring a write.

## A prediction in prose is what failed, so the fix is a test

`ApprovalSurfaceInventoryTest` holds three lists: which modules register the approval actions and under which `access`; which classes under `Classes/Controller` reach `AgentRuntime::approve()`/`submitInput()`; and that every AJAX route on `resumeAction`/`submitInputAction`/`approveAction` opens with `denyNonAdmin()`. It then requires ADR-130 to **name each module the scan finds**, matched as a whole identifier.

That matching detail is not cosmetic. An earlier draft used `assertStringContainsString`, and a module named `nrllm_run` satisfied it through the existing `nrllm_runs` — a substring collision that let a fourth surface through. A second control found worse: a fourth module under a *different* controller was invisible to all three assertions and the suite stayed fully green. Both holes are closed; each guard was then seen to fail on the defect it exists for.

**The ADR states what the three lists do not cover**, because they do not cover everything: a class outside `Classes/Controller`, exposed under a fourth action name, still passes. That was tried and it stayed green. Saying so is the point — the check exists to make this record's enumeration falsifiable, not to become a second registry. An earlier draft of this very paragraph also planted the literal `nrllm_run` into the text as an example, which would itself have satisfied the naming check for a module of that name; no example is given now, deliberately.

## Citation rot, found and fixed

Eleven line-anchored citations in ADR-169 and ADR-171 pointed at ranges this amendment moved. One predated it: **ADR-171 cited `ResumeCoordinator.php:204` for the `mayActOnRun` check, and that line is blank** — the four-eyes commit inserted a guard above it four days ago. The personas record rotted within a day of merging.

All eleven are recomputed. The general problem is filed as #793: 74 such citations exist and `AdrLifecycleTest` checks `:ref:` targets but not these.

## Related Issue

Closes #787 — whose own framing this change corrects; the correction is a comment on the issue.

Also filed while sweeping: #791 (ADR-119 counts twelve submodules, there are fourteen plus one outside), #792 (ADR-108/094 count 40/41 builtins, there are 46), #793 (line-anchored citations rot unchecked).

## Type of Change

- [x] Documentation update
- [x] New test (a mechanical gate for this defect class)

## Deliberately not done

ADR-130 **constraint 4** is expired too — `tasks_manage` "does not exist yet", and the editing module shipped without it. Its amender is ADR-169, which is `:Status: Proposed`. The lifecycle forbids a proposal writing `:Amends:`, so that amendment is owed by whichever change accepts ADR-169. Pre-empting it would break the same rule in the other direction.

## Gates

`ApprovalSurfaceInventoryTest` OK (4 tests, 79 assertions), `AdrLifecycleTest` OK (6 tests, 1187 assertions), all three repo checks exit 0.

`rector -n` at the CI-pinned PHP 8.2 did not run — `.Build` here is resolved at 8.4 and `platform_check.php` fatals before it starts. One new PHP file reaches CI without that gate having run locally.

## Checklist

- [x] I have added a test that proves the gate works, and watched each assertion fail
- [x] `CHANGELOG.md` entry under the existing `### Changed`
- [x] Both ends of the amend link written, verified by `AdrLifecycleTest`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,28 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- **ADR-130's constraint 3 is amended by ADR-131** (`#787`). The record called
+  `agent_approve` doubly unreachable for non-admins because it named two
+  approval surfaces and both were admin-gated. Both still are: `nrllm_runs` is
+  `access => 'admin'` and the Playground's resume path still calls
+  `denyNonAdmin()`. The enumeration was what went out of date — the editor
+  module `nrllm_aitasks` is `access => 'user'` and registers `approve` and
+  `submitInput`, so a non-admin whose group has that module ticked and who holds
+  the grant decides other users' runs today (both switches are required). The
+  amendment states the three bounds that apply on top of module access — the
+  module tick is required but bounds nothing on its own — in the order
+  `ResumeCoordinator::approve()` evaluates them: `mayActOnRun()`, the opt-in
+  four-eyes refusal (ADR-172), and the approver tool gate (ADR-133) that
+  withholds a write the approver could not run themselves. That list is the
+  approve path; `submitInput` is recorded next to it with the same first and
+  last gate and no four-eyes check (ADR-150). The stale citations ADR-169 and
+  ADR-171 carried into the same records are corrected with it. A new
+  `ApprovalSurfaceInventoryTest` pins which modules register `approve` /
+  `submitInput` under which `access`, which classes under `Classes/Controller`
+  reach the approval runtime, and that ADR-130 names each module it finds — so a
+  new approval module cannot be registered without editing the record. The ADR
+  states what the three lists do not cover. No behaviour changes.
+
 - **One MCP operation now has one timeout, not one per HTTP request** (`#773`,
   ADR-170). A tool call against an MCP server is three requests — the protocol
   handshake, its confirmation, then `tools/call` — and each carried a full
--- a/Documentation/Adr/Adr130BackendUserGrants.rst
+++ b/Documentation/Adr/Adr130BackendUserGrants.rst
@@ -4,8 +4,10 @@
 ADR-130: Capability grants for backend users
 ===========================================
 
-:Status: Accepted
+:Status: Accepted (constraint 3 enumerated the approval surfaces and the
+   enumeration was not exhaustive — see :ref:`ADR-131 <adr-131>`)
 :Date: 2026-08-06
+:Amended: 2026-08-06 by :ref:`ADR-131 <adr-131>`
 
 Context
 =======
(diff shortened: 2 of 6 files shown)


CybotTM:
**Review record.** Copilot is out of review quota — monthly and account-wide, now recorded as such, so no PR this month gets a bot review and re-requesting cannot change it. This was reviewed without one.

**The issue this closes had the wrong premise, and correcting it is the finding.** #787 said ADR-130's constraint expired. It did not: both surfaces it names are still admin-only (`Modules.php:299-301`, `:280-282`, plus `denyNonAdmin()` at `ToolPlaygroundController:295`/`:381`). A *third* surface arrived. The correction is on the issue.

**Which record made it reachable was established three ways**, not assumed: `git log -S` on two separate symbols, and ADR-131's own Consequences, which already said it in prose. Only the two lifecycle fields were missed.

**The bound order in my own brief was wrong and the code overruled it.** I specified `mayActOnRun → ADR-133 → ADR-172`. `ResumeCoordinator::approve()` evaluates `mayActOnRun` (`:205`), then ADR-172's self-approval refusal (`:222-226`), then ADR-133's `approverRefusal()` (`:314`). The record states the runtime order and says so.

**A prediction in prose is what failed, so the fix is a test.** That test then had to survive its own negative controls, and it did not at first:

- a fourth module named `nrllm_run` passed the naming assertion, because `assertStringContainsString` matched the existing `nrllm_runs` — a substring collision;
- a fourth module under a *different* controller was invisible to all three assertions and the suite stayed fully green.

Both closed: whole-identifier matching, the controller filter dropped, and a new assertion pinned to the classes that actually call `AgentRuntime::approve()`/`submitInput()` — a definition a registration cannot dodge. Each guard was then seen to fail on the defect it exists for.

**What the test does not cover is written into the ADR**, because it does not cover everything: a class outside `Classes/Controller`, exposed under a fourth action name, still passes. That was tried and it stayed green. An earlier draft of that very paragraph also planted the literal `nrllm_run` into the text as an example — which would itself have satisfied the naming check for a module of that name. No example is given now.

**Citation rot, found and fixed.** Eleven line-anchored citations in ADR-169 and ADR-171 pointed at moved ranges. One predated this branch: ADR-171 cited `ResumeCoordinator.php:204`, and that line is blank — the four-eyes merge inserted a guard above it. The personas record rotted within a day of merging. Filed generally as #793.

**Rector was the one gate I could not run locally all session, and it was right.** I built a second `.Build` resolved at PHP 8.2 rather than keep reporting the gate as unrun: it wanted `assertMatchesRegularExpression` in place of `assertSame(1, preg_match(...))`, which also prints the pattern on failure. Applied, then `cgl` and both dry runs re-verified in that order.

**Gates:** `ApprovalSurfaceInventoryTest` OK (4 tests, 79 assertions), `AdrLifecycleTest` OK (6 tests, 1187 assertions), `phpstan` level 10 clean, `cgl -n` clean, `rector -n` clean at the CI-pinned 8.2, all three repo checks exit 0.

PR #801, opened 2026-08-18
feat(runs): say when a run was released by the person who started it

### LINKED ISSUE #785: A self-approved run is not distinguishable from a reviewed one
An approval event records who decided, and the run records who started it, but nothing puts the two side by side. So a run approved by the person who started it is indistinguishable, at a glance, from one a second person reviewed.

`AiActorContext::mayActOnRun()` ends with `return $this->backendUserUid > 0 && $this->backendUserUid === $run->beUser;` — the run's own initiator may act on it. Self-approval is therefore not an edge case that slips through: it is the **fallback path** for anyone who is not an administrator, not a service account, and holds no `AGENT_APPROVE` grant. On a single-user installation it is the only path there is.

That is the right behaviour. Refusing it would leave the run suspended forever and make the extension unusable wherever one person does everything, which is most small installations. The problem is not that it happens.

**The problem is that the record does not say it happened.** `ResumeCoordinator` calls `recordApproval($handle, $decision->approved, $decision->decidedByBeUser)`, the payload is `{approved, decidedBy}`, and `decidedBy` is carried into the run timeline. The data is complete. But a repo-wide search finds no reader that compares `decidedBy` against the run's `beUser`, and no surface that labels the result. An auditor would have to hold two uid numbers in mind and notice they match.

The value of an approval gate is not that two people looked. It is that the record states truthfully how many did. On this one point it currently does not.

**What closing it needs**

A derived, displayed distinction wherever an approval is shown — the approvals inbox, the run timeline, the governance readout: this decision was made by the initiator, or by someone else. No new column is needed; both values are already stored.

Worth deciding at the same time, because it is the same question one step further: whether a self-approval should be *visible* only, or whether an installation that has separated the roles should be able to *refuse* it. The second is a policy switch and needs its own argument — it must not become the default, or the single-user case breaks.

Related: this is the sharpest instance of a general question the persona work is circling — what the system does when two roles it was designed to separate turn out to be one person. Refusing breaks small installations; allowing silently makes the audit overstate the control; allowing visibly is the only option that survives both.

Ref: ADR-084, ADR-132, ADR-133, ADR-150, ADR-101.

CybotTM:
## Description

An approval event records who decided. The run records who started it. Nothing put the two side by side, so a run released by its own initiator read exactly like one a second person reviewed.

**Self-approval is not an edge case.** `AiActorContext::mayActOnRun()` ends with `$this->backendUserUid > 0 && $this->backendUserUid === $run->beUser`, so it is the *fallback* path for anyone who is not an administrator, not a service account and holds no `agent_approve` grant — and on a single-user installation it is the only path there is.

ADR-172 (#789) shipped `require_second_approver`, a per-configuration switch that refuses it. That is the enforcement half and it is done. **This is the visibility half, and #789 makes it more valuable rather than less:** the switch is off by default, so an operator can believe four eyes applied to a configuration where the same person did both jobs.

## Two ways of not knowing are two labels

A first draft collapsed them into one, and it was false for half the cases. `AgentRun::$beUser` is `0` for every run a service account or non-BE caller started. Approve such a run as backend user 5 and the record stores `decidedBy=5` — the timeline even prints it. "Approved, but the record does not say by whom" is then simply wrong: the record does say. It does not say who *started* it.

So there are four states, not three:

| | |
| --- | --- |
| **Self-approved** | both uids present and equal |
| **Approved by someone else** | both present, different |
| **Starter unrecorded** | a service-account run released by a backend user |
| **Decider unrecorded** | an approval whose decider was not stored |

Only granted approvals carry a label. A denial carries none, because ADR-172 lets an initiator deny their own run on purpose.

## Where the two surfaces can differ, stated rather than papered over

The comparison is derived once, in `ApprovalAttribution`, and read by the approvals inbox, the run timeline and the terminal-run list. The timeline states **each** fence; the list **collapses** several by rank. On a multi-fence run they can therefore read differently — a run with fences by user 5 and by an unrecorded decider shows "the record does not say by whom" in the list while the timeline prints `decidedBy=5` beside it.

That is recorded on the enum as an accepted price, and the docblock that claimed the two surfaces "cannot disagree" is corrected. Neither surface owns the rule; the list additionally collapses.

## What this does not do

**No policy changes.** Refusing a self-approval is ADR-172's job and it exists. This makes what happened legible, nothing more.

## Related Issue

Closes #785

## Type of Change

- [x] New feature (non-breaking)
- [x] Documentation update

## Corrections made during review, worth naming

- The `string` type on two view DTOs was justified with "the view appends it to a translation key". Neither view does — the partial is an `f:switch` over four literal keys, and a named test in this branch asserts that no surface assembles a key. Three places said the opposite of a green test; all three now describe the design that shipped rather than the one that was rejected.
- The empty-cell explanation enumerated three inputs and a fourth reaches it: a granted approval whose payload JSON does not decode is dropped by the fail-soft read. Widened to "loaded or decoded" in all four places it appears.

## Gates

`cgl -n`, `phpstan` level 10, `unit` (7158 tests) and functional scoped to `--filter 'Approval|AgentRun'` (68 tests) all exit 0, plus all three repo checks.

**`rector -n` at the CI-pinned PHP 8.2 did not run** — `.Build` here is resolved at 8.4 and `platform_check.php` fatals before it starts.

**Not checked in a browser.** The new Approval column carries four full-sentence labels in a seven-column table. Nobody has looked at it at 1440px. The judgement to prefer sentences over one-word chips was taken for legibility and is worth overturning if it wraps badly — that is a review call, and it is stated here rather than left to be discovered.

## Note on ADR numbering

This record is ADR-173. A sibling branch (routing instrumentation) independently claimed 173 and has been renumbered to 174 — both were told to "confirm the next free number" rather than being given disjoint ranges, which is the parallel-agent collision that rule exists to prevent.

## Checklist

- [x] Tests prove the feature works, and the central one was seen to fail without it
- [x] Documentation updated
- [x] `CHANGELOG.md` entry under an existing heading
- [x] ADR-173 added and listed in `Documentation/Adr/Index.rst`
- [x] EN and DE labels for every new string

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [Unreleased]
 
 ### Added
+- **A self-approved run says so** (`#785`). An approval granted by the backend user who started the run is now labelled as such — on the run timeline, and in the *Recent runs* table of the Agent Runs inbox, which gains an *Approval* column. An approval granted by anybody else is labelled too, so the two are told apart at a glance instead of by comparing two uid numbers.
+
+  Derived at read time from values that were already stored, so it is right for runs recorded before this change. The readout compares the two uids in one place, `Domain\Enum\ApprovalAttribution`, and both surfaces render its answer; ADR-172's four-eyes gate keeps its own, stricter comparison in `AiActorContext::isInitiatorOf()`.
+
+  Granted approvals only: a denial carries no label, because ADR-172 lets an initiator deny their own run on purpose. Nothing reads *self* unless both uids are present and equal, and the two ways one can be missing are two labels rather than one: a run started by a service account and released by a backend user reads *Approved, but the record does not say who started the run*, while an approval whose decider was not recorded reads *Approved, but the record does not say by whom*. An empty cell is not the opposite of any of that — it means the run passed no fence, was denied, or its approvals could not be loaded or decoded.
+
+  This is a readout and refuses nothing. Whether self-approval is allowed remains the `require_second_approver` switch (ADR-172). See ADR-173.
+
 
 - **Per-call cost and real token counts on the telemetry row** (`#770`). Every
   provider call now records what it actually consumed — input tokens, output
--- a/Classes/Controller/Backend/AgentRunController.php
+++ b/Classes/Controller/Backend/AgentRunController.php
@@ -64,6 +64,14 @@ use TYPO3\CMS\Extbase\Mvc\Controller\ActionController;
  * has no grant equivalent, so the approval grant widens the list but not the
  * detail page (ADR-153).
  *
+ * The list therefore carries ONE fact the detail page would refuse to that same
+ * viewer: the ADR-173 approval attribution, derived from the approval events of
+ * runs a grant holder may list but not open. Deliberate and bounded — it is the
+ * same audit-only class as the `decidedBy` uid above — four labels and an empty
+ * cell, and it names nobody. The
+ * alternative, showing the column only on the viewer's own rows, would hide
+ * exactly the rows an approval grant exists to review.
+ *
  * The page works fully with JavaScript OFF: native `<f:form>` POST, a
  * POST-redirect-GET flush with session flash messages, and a 422 in-place
  * re-render that preserves the operator's raw input. The CSRF defence is the
@@ -286,7 +294,7 @@ final class AgentRunController extends ActionController
 
         $this->moduleTemplate->assignMultiple([
             'waiting'        => $this->viewFactory->buildWaiting($waitingRuns ?? [], $viewer instanceof BackendUserAuthentication ? $viewer : null),
-            'terminal'       => $this->viewFactory->buildTerminal($terminalRuns ?? [], $actor),
+            'terminal'       => $this->viewFactory->buildTerminal($terminalRuns ?? [], $actor, $this->approvalDeciders($terminalRuns ?? [])),
             'dataLoadError'  => $dataLoadError,
             'errorRunUuid'   => $errorRunUuid,
             'rawInput'       => $rawInput,
@@ -295,6 +303,33 @@ final class AgentRunController extends ActionController
         ]);
 
         return $this->moduleTemplate->renderResponse('Backend/AgentRun/List');
+    }
+
+    /**
+     * Who granted each listed run's approvals (ADR-173), in ONE statement for the
+     * whole table — the alternative, an event read per row, would be twenty
+     * queries for a marker.
+     *
+     * No authorisation is added or needed here: the rows are exactly the ones
+     * the list query already released to this viewer. That is not an actor-scoped
+     * set — for an admin or an `agent_approve` grant holder `$restrictTo` is null
+     * and the list is every user's run on purpose (ADR-131), which is the case
+     * this marker exists for. A `decidedBy` uid is audit metadata the inbox
+     * docblock already names as such.
+     *
+     * @param list<AgentRun> $runs
+     *
+     * @return array<int, list<int>>
+     */
+    private function approvalDeciders(array $runs): array
+    {
+        if ($runs === []) {
+            return [];
+        }
+
+        return $this->persister->findApprovalDeciders(
+            array_map(static fn(AgentRun $run): int => $run->uid, $runs),
+        );
     }
 
     /**
--- a/Classes/Domain/Enum/ApprovalAttribution.php
+++ b/Classes/Domain/Enum/ApprovalAttribution.php
@@ -1,0 +1,190 @@
+<?php
+
+/*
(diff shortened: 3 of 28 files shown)


CybotTM:
**Review record.** Copilot is out of review quota for the month, account-wide. No bot review on this or any PR until the reset; re-requesting cannot change it. Reviewed without one.

**The first draft's label was factually wrong, and a verifier caught it by construction rather than by reading.** `ApprovalAttribution::fromDecision()` returned UNRESOLVED whenever *either* uid was `<= 0`. But `AgentRun::$beUser` is `0` for every run a service account started, so approving such a run as backend user 5 stored `decidedBy=5` — and the label said "the record does not say by whom". The record does say. It does not say who *started* it. Two facts, now two labels.

**Where the two surfaces disagree is written down instead of denied.** A docblock claimed the list and the detail page "cannot disagree about who released a run, because neither owns the rule". The second half is true; the first is not. The timeline states each fence, the list collapses several by rank in `fromDecisions()`, so on a multi-fence run they read differently — a run with fences by user 5 and by an unrecorded decider shows "does not say by whom" in the list while the timeline prints `decidedBy=5` beside it. That is now stated as the accepted price it is.

**Three places described a design that was rejected.** Two view DTOs justified their `string` type with "the view appends it to a translation key", and a test was named `…SuffixesTheViewsAppend()`. No view appends anything: the partial is an `f:switch` over four literals, its own header says an assembled key was rejected for having no `default`, and a green test in this branch asserts that neither surface assembles one. All three now describe what shipped.

**An enumeration that read exhaustive was one short.** The empty-cell explanation named three inputs; a fourth reaches it — a granted approval whose payload JSON does not decode is dropped by the fail-soft read. Widened in all four places rather than left approximate.

**Rector, finally runnable.** I built a second `.Build` at PHP 8.2 instead of continuing to report this gate as unrun. It wanted a local helper called on `$this` rather than statically. Applied, `cgl` after it, both dry runs clean.

**Gates:** `cgl -n`, `phpstan` level 10, `unit` (7158 tests), functional `--filter 'Approval|AgentRun'` (68 tests), `rector -n` at 8.2, all three repo checks — all exit 0.

**Not checked in a browser, and this is the one thing a reviewer should look at.** Four full-sentence labels in a seven-column table at 1440px is unmeasured. The judgement to prefer sentences over one-word chips was taken for legibility; if it wraps badly, overturn it. Stated here rather than left to be discovered.

**ADR numbering:** this is 173. A sibling branch independently claimed 173 and is renumbered to 174 — both were told to confirm the next free number instead of being given disjoint ranges, which is exactly the parallel-agent collision that rule exists to prevent.

CybotTM:
Rebased onto `main` after #800 merged, now `0c1d2743`.

`CHANGELOG.md` auto-merged cleanly — #800's entry sits under a different heading — but a conflict-free auto-merge can still drop lines neither side deleted, so it was checked rather than assumed: all 10 of `main`'s `[Unreleased]` bullets are present, the three headings are unchanged, and no bullet appears twice.

The review record above applies unchanged, and that is verified rather than asserted: the change set is `28 files changed, 1421 insertions(+), 31 deletions(-)` — byte-for-byte the same counts as before the rebase, so nothing in the reviewed code moved.

Unit suite on the rebased tree: 7163 tests, exit 0. It now also runs #800's new `ApprovalSurfaceInventoryTest` alongside this branch's work.

Next in the sequence: #802 rebases after this one merges, which also closes the 172 → 174 gap its toctree currently carries.

CybotTM:
Rebased onto `main` after #802 merged, now `5a9dba70`.

Both `CHANGELOG.md` and `Documentation/Adr/Index.rst` conflicted, as expected — #802 appends its toctree line at the same position this one does. **Rebuilt rather than merged**, for the reason this repository already learned the hard way: "ours holds only my additions" is right for the first merge of a series and wrong for every later one. Both files were reset to their `main` version verbatim and this branch's block re-inserted under the matching heading, with the toctree line placed in ascending position.

Verified after, not assumed:

- three `### ` headings in `[Unreleased]`, no bullet appearing twice, no real conflict markers (matched as `^<<<<<<< ` / `^>>>>>>> ` with the trailing space — a bare `=======` is RST underlining and a naive grep reports it as a conflict);
- the toctree reads `… Adr172, Adr173, Adr174` and every `Adr*.rst` on disk is listed;
- `check-changelog-unreleased.php` passes.

**The review record above still applies, and that is checked rather than claimed:** the change set is `28 files changed, 1421 insertions(+), 31 deletions(-)` — the same counts as before the rebase, so nothing in the reviewed code moved.

Unit suite on the rebased tree: 7197 tests, exit 0. That is 60 more than before the rebase, which is #802's own tests now running alongside this branch's.

One correction to a check I ran while doing this: my first "is every ADR in the toctree" query used `sed 's/.rst//'` with an unescaped dot, which cut names mid-word and reported three records as missing. They were not. Re-run with the dot escaped and anchored: none missing.

Issue #772, opened 2026-08-13
No per-call quality signal: ADR-156's second criterion is unmeasurable


CybotTM:
There is no per-call quality signal, which means no online experiment — canary or otherwise — can be evaluated today.

What exists is `tx_nrllm_eval_result`: golden-set runs, out of request, keyed by (set identifier, model, grader), read as a per-**model** mean (`EvaluationResultRepository.php:98-130`), produced by a CLI command nothing schedules. That supports a statement like "model A scores better than model B on the golden set". It cannot support "this canary request was as good as its control counterpart", because the value is constant within a model and therefore identical for both arms.

Nothing else fills the gap. Governance events record blocks only, never allows, so there is no denominator. There is no rating, no thumbs, no acceptance signal, no feedback of any kind. Retries are not counted.

Any promotion rule phrased as a quality delta would be a rule over a metric nobody records — the declaration-nothing-reads failure ADR-142 and ADR-156 were written to avoid.

**What closing it needs**

A request-level outcome contract hanging off `correlation_id`.

**Not** to be pre-decided: that approval equals quality. An approval can be withheld for governance, policy or content reasons that say nothing about whether the model did well. Treating a refusal as a quality signal would poison the metric with governance behaviour.

The likely shape is an explicit consumer feedback signal or a use-case-specific outcome measurement, defined deliberately rather than inferred from an existing gate.

Precondition for ADR-156's second activation criterion, and for any canary evaluation.

Ref: ADR-156, ADR-142.

PR #802, opened 2026-08-18
feat(telemetry): record what a call cost and what it looked like before routing

### LINKED ISSUE #770: No per-call cost or real token counts: telemetry and cost cannot be joined

### LINKED ISSUE #771: Complexity is measured after model selection, so it cannot inform it
The complexity score is measured **after** the model has been chosen, and partly **from** the chosen model. It is therefore sound post-decision observability and cannot become an input to the same decision.

`resolveModel()` runs before `fitToContextWindow()` on every path (`LlmServiceManager.php:858→871`, `1375→1383`), and `measureComplexity()` is called from the fit. The size term of the score is estimated tokens against the budget of the model that was already selected — context utilisation is by definition a property of the chosen window. The decision path never sees the messages at all: `decide(array $criteria, ?RoutingPolicyMode)` takes criteria, not a payload.

ADR-156 says as much: nothing routes on this.

Any adaptive-routing step — bucketing requests before the decision, letting complexity act as a soft ranking signal — rests on a data flow that does not exist. That is a prerequisite, not a detail of the ranking change.

**What closing it needs**

A model-independent request fact set, formed before model selection:

- operation;
- message and turn count;
- tool count;
- payload bytes;
- optionally a provider-independent token estimate;
- request shape.

Explicitly **not** in it: context utilisation, the chosen model, its window, its price.

The existing post-fit complexity record stays as it is — it answers a different question and both are worth having.

Precondition for anything adaptive. Pairs with the per-call cost gap and the missing online outcome signal.

Ref: ADR-156.

CybotTM:
## Description

Everything adaptive rests on two measurements the extension did not have. **Neither needed production data to build** — they need it to *evaluate*, and conflating those two is why this looked blocked when it was not.

## #770 — per-call cost and real tokens

`tx_nrllm_telemetry` carried routing, success, latency and a complexity estimate but no money and no actual counts. `routing_signal_cost` is a ranking signal, not a price; `complexity_tokens` is explicitly a guess. Real cost lived in `tx_nrllm_service_usage` as a **daily aggregate with no `correlation_id`**, so no cost could be joined to a call, a request shape or a complexity bucket.

Five columns now carry it under the id the call already has: `actual_input_tokens`, `actual_output_tokens`, `actual_cost`, `response_model`, `provider_retries`.

**`NULL`, never `0` — and that distinction is the defect this closes, not a nicety.** `Model::hasPricing()` is false when both prices are zero, so Ollama, Groq and every local model recorded `estimated_cost: 0`, indistinguishable from *unpriced* — on precisely the arm a cheap-model experiment is about. `actual_cost` is `NULL` there instead.

Token columns are `NULL` where the provider reported no usage block. `UsageStatistics` types its counts as plain `int`, so the only available signal is **all three zero at once**; a surviving `0` is therefore a measurement, not a gap. Making that type nullable was the alternative and is a breaking `@api` change — named and rejected in the ADR rather than done quietly.

## #771 — pre-routing request facts

The complexity score is measured **after** the model is chosen and partly **from** it: `resolveModel()` runs before `fitToContextWindow()`, and the size term is estimated tokens against the budget of the model already selected. The decision path never sees the messages at all — `decide()` takes criteria, not a payload.

So a second, model-independent fact set is formed **before** selection: message and turn count, tool count, payload bytes, a provider-independent token estimate, request shape.

Deliberately out: context utilisation, the chosen model, its window, its price. Those are exactly what made the first measurement unusable as an input.

**The post-fit record stays.** The two answer different questions, and the ADR states where they diverge rather than calling them equivalent: the system prompt is never the difference between them, **pruning is**, and on a row where the fit dropped turns the pre-routing counts are the larger of the two.

## Nothing routes on any of it

ADR-174 keeps ADR-156's observer-only status intact. This supplies the evidence its three activation criteria need instead of pre-empting them. Two of those criteria were unmeasurable before this change; the third — an online quality signal — still is, and #772 is where that lives.

## Verified against real traffic, with the limits stated

The instrumented path was driven against the seeded local Ollama with deliberately varied turn counts, tool presence and payload sizes, and the recorded rows were read back rather than assumed.

That proves the instruments write real values end to end. It produces **no representative cost data** — a local model is priced at zero, which is the very case `NULL` now distinguishes — and **no quality data at all**.

## Related Issue

Closes #770, closes #771

## Type of Change

- [x] New feature (non-breaking)
- [x] Documentation update
- [x] Schema change (`ext_tables.sql`)

## Known limits, named rather than buried

- **Streaming retry attribution.** `ProviderRetryCounter` is process-wide, and `drain()` is a generator: between the capture and the `finally`, the consumer's own code runs, so retries it causes land on this stream's `provider_retries`. Scoped in the ADR rather than justified with reasoning that does not hold for a generator.
- **`provider_retries IS NULL`** was documented as a meaningful state that no production write site can produce. Corrected.
- The wiring of the shared counter through three constructors is now pinned by a functional test rather than working by luck.

## Gates

`cgl -n`, `phpstan` level 10, `unit` (7166 tests) and functional scoped to `--filter 'Telemetry|Complexity|RequestFacts'` (37 tests) all exit 0, plus all three repo checks.

**`rector -n` at the CI-pinned PHP 8.2 did not run** — `.Build` here is resolved at 8.4 and `platform_check.php` fatals before it starts.

## Note on ADR numbering

This record was written as ADR-173 and is renumbered to **174**: a sibling branch independently claimed 173. Both were told to "confirm the next free number" instead of being given disjoint ranges — the parallel-agent collision that rule exists to prevent. The toctree jumps 172 → 174 until the sibling merges.

## Checklist

- [x] Tests prove the feature works
- [x] Documentation updated
- [x] `CHANGELOG.md` entry under existing headings
- [x] ADR-174 added and listed in `Documentation/Adr/Index.rst`
- [x] Schema change carries its `ext_tables.sql` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,52 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- **Per-call cost and real token counts on the telemetry row** (`#770`). Every
+  provider call now records what it actually consumed — input tokens, output
+  tokens, the model the provider named on the response, the cost derived from
+  its pricing, and how many HTTP attempts had to be repeated — under the
+  correlation id the call already carries.
+
+  `tx_nrllm_service_usage` holds the same figures as a daily aggregate with no
+  per-call key, so nothing there can be joined to a call, a request shape or a
+  complexity bucket. That made ADR-156's third activation criterion, "real cost
+  drops", not computable rather than merely awkward.
+
+  `NULL`, never `0`, where nothing was measured. A model priced at zero —
+  Ollama, Groq, anything local — used to record a cost of `0`, which is also
+  what an unpriced model records; it now records real tokens and no cost. Token
+  counts are `NULL` where the provider reported no usage block, and a recorded
+  `0` is a measured zero. Streamed runs record no token counts at all: they
+  carry an estimate, and an estimate does not belong in a column named
+  `actual`. See ADR-174.
+
+  Additive on the `@api` surface: `ProviderCallUsage`,
+  `TelemetrySignals::recordCallUsage()`, and an optional trailing
+  `ProviderRetryCounter` argument on `ProviderAdapterRegistry`'s constructor.
+
+- **Request facts, measured before the model is chosen** (`#771`). Message
+  count, turn count, tool count, payload bytes, a provider-independent token
+  estimate and the request shape are now recorded for every
+  configuration-driven send, formed before the pipeline runs — which is before
+  any model resolution.
+
+  The counts describe the transcript the caller sent. The configuration's system
+  prompt is prepended later, from options built out of the resolved model, so it
+  is in none of them — the same line the complexity record draws, which is what
+  keeps the two comparable.
+
+  The existing complexity record is measured *after* the model is chosen and
+  partly *from* it: its size term is estimated tokens against the budget of the
+  model already selected. It stays as it is and answers a different question;
+  the new set is model-independent by construction and deliberately excludes
+  context utilisation, the chosen model, its window and its price.
+
+  Nothing routes on either set. ADR-156 keeps its observer-only status and its
+  three activation criteria unchanged. See ADR-174.
+
+  Additive on the `@api` surface: `RequestFacts` and
+  `TelemetrySignals::recordRequestFacts()`.
+
 - **Four-eyes approval, per configuration** (`#786`). A new configuration field,
   *Require a second approver*, refuses an approval from the backend user who
   started the run. Default off, so every existing record keeps the
--- a/Classes/Domain/ValueObject/ProviderCallUsage.php
+++ b/Classes/Domain/ValueObject/ProviderCallUsage.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * Copyright (c) 2025-2026 Netresearch DTT GmbH
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+declare(strict_types=1);
+
+namespace Netresearch\NrLlm\Domain\ValueObject;
+
+/**
+ * What one provider call actually consumed, and what that cost (ADR-174).
+ *
+ * The counterpart of {@see RequestComplexity}: that one estimates a request
+ * before it is answered, this one reports what came back. Both hang off the
+ * same telemetry row, so a shape, a decision and a price share one
+ * ``correlation_id`` — which is what ``tx_nrllm_service_usage`` cannot give,
+ * being a daily aggregate with no per-call key.
+ *
+ * **Null is a measurement that did not happen, and never a zero.** A provider
+ * that reports no usage block leaves the token fields null; a model with no
+ * pricing leaves the cost null. Both cases previously surfaced as ``0`` — the
+ * cost of a free local model and the cost of an unpriced one were the same
+ * number, which is precisely the arm a cheap-model experiment is about.
+ *
+ * Prompt-free like the row it is written to: two counts, a price and a model
+ * name.
+ *
+ * @api
+ */
+final readonly class ProviderCallUsage
+{
+    /**
+     * @param ?int   $inputTokens   prompt tokens as the PROVIDER reported them, or null where it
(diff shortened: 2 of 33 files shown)


CybotTM:
**Review record.** Copilot is out of review quota for the month, account-wide. Reviewed without one.

**The premise of "not buildable" was mine and it was wrong.** I had reported these instruments as blocked on production data. They are not: the data is needed to *evaluate*, not to *build*. Both are built here, and the instrumented path was driven against the seeded local Ollama with the recorded rows read back rather than assumed.

**The `NULL`-versus-`0` distinction is the defect, not a nicety.** `Model::hasPricing()` is false when both prices are zero, so every local and zero-priced model recorded `estimated_cost: 0` — indistinguishable from *unpriced*, on precisely the arm a cheap-model experiment is about. `actual_cost` is `NULL` there now.

The token columns were harder and the reasoning is worth checking: `UsageStatistics` types its counts as plain `int`, so the only signal for "the provider reported nothing" is **all three zero at once**, and a surviving `0` is therefore a measurement. Making that type nullable was the clean alternative and is a breaking `@api` change — named and rejected in the ADR rather than done quietly.

**Corrections made during review, each a claim that did not hold:**

- Streaming retry attribution was justified with reasoning that does not survive generator suspension — `drain()` yields, and the consumer's own retries land on this stream's counter between the capture and the `finally`. Scoped rather than justified.
- `provider_retries IS NULL` was documented as meaningful in three places and no production write site can produce it.
- A test named for prompt-freedom was tautological: the record it built contained no request content at all, so no production change could have failed it.
- The ADR contradicted itself about whether `operation` is part of the fact set — it is not, and `RequestFacts` has six properties with no operation among them.
- The two token figures were called a like-for-like pair. They are not on a pruned row: the system prompt is never the difference between the counts, **pruning is**, and the pre-routing byte count is the larger where the fit dropped turns.

**Rector's one production-code change needed a decision, not an application.** It wanted a redundant `callUsage: null` removed — correct about the redundancy, but the argument carried a comment explaining why a streamed response deliberately records no usage. Deleting both would have removed the reasoning with the redundancy. The argument is gone; the reasoning stays as a comment in the same place.

**Nothing routes on any of this.** ADR-156 keeps its observer-only status. Two of its three activation criteria become measurable here; the third — an online quality signal — does not, and #772 is where that decision lives.

**Gates:** `cgl -n`, `phpstan` level 10, `unit` (7166 tests), functional `--filter 'Telemetry|Complexity|RequestFacts'` (37 tests), `rector -n` at the CI-pinned 8.2, all three repo checks — all exit 0.

**ADR numbering:** written as 173, renumbered to **174** across 32 files after a sibling branch independently claimed 173. The toctree jumps 172 → 174 until that sibling merges.

Issue #691, opened 2026-08-10
Management grant — needs a management surface first


CybotTM:
`BackendUserGrant` has exactly two cases today: `TASKS_USE` and `AGENT_APPROVE`. Both exist because ADR-130 requires a grant to be added *together with its consumer* — there is no wildcard grant and no grant without an enforcement point.

A management surface (configuring providers, models, configurations and tasks outside the admin-only backend module) has no grant, because the surface does not exist. Someone who should manage nr-llm today must be a TYPO3 administrator.

**Why this is not just "add an enum case".** Adding `MANAGE` before a management surface exists would produce exactly the failure ADR-130 was written to avoid: a checkbox in the backend group form that reads as an access control and gates nothing. ADR-023 shipped that once and it had to be removed in 0.26.0 (ADR-117).

**Order of work:** the surface first, the grant with it.

Ref: ADR-130 (`Adr130BackendUserGrants.rst`), ADR-131 (`Adr131EditorModule.rst`), ADR-117, ADR-023.

CybotTM:
Answered rather than built, by ADR-169 section 5 — accepted in #813.

This issue said the surface had to come first and the grant with it. The audit that followed found the order was not the problem: after ADR-169 section 2 settles which records a non-admin may manage (`tx_nrllm_task`, `tx_nrllm_promptsnippet`, `tx_nrllm_configuration`) and section 4 settles that they are managed through FormEngine with `exclude => true` on the governance and spend fields, both halves are authorised by `tables_modify` and `non_exclude_fields`. Those are TYPO3's own permissions on the same fields. A `MANAGE` case would sit in front of nothing — which is the failure this issue was written to avoid, arrived at from the other direction.

The premise that made a grant look necessary was also wrong, and it is worth stating because it appears in three places in this repository's history: `rootLevel => -1` is `TYPE_BOTH`, not root-only. A `tx_nrllm_*` record stored on a page inside a web mount is editable by a non-admin holding `tables_modify` **today**, with no TCA change and no module. What was missing was never the grant; it was the field boundary, which section 4 supplies.

The reservation is retired in all three places that carried it: ADR-130 named constraint 4, ADR-131's "what stays out" bullet, and `BackendUserGrant`'s class docblock. The docblock says not to add the case back without naming an action those two permissions do not already cover.

Two things did **not** close with this and are filed:

[...]

PR #804, opened 2026-08-18
docs(roadmap): the last known gap closed in code

### LINKED ISSUE #776: ROADMAP.md still lists the closed #731 as open
`ROADMAP.md:23` still lists `#731` — "a per-call injected snippet has no trust-zone ceiling" — as open work. `#731` is closed, and so is its follow-up `#761`.

The file is the first place a reader looks for "what is not built", so a closed item sitting in it is worse than a gap: it sends someone to re-derive a decision that has already shipped as ADR-164 and ADR-165.

One-line edit. Worth a quick pass over the neighbouring entries in the same change — if one has gone stale, others may have too.

CybotTM:
ROADMAP.md listed #731 — "a per-call injected snippet has no trust-zone ceiling" — as open work under **Known gaps**. It is closed, and closed by code rather than by a decision: ADR-164 binds a run's forced sources against the same trust ceiling as a configuration's own, and ADR-165 does it again on resume. That distinction is what the section is about, so the entry had to be checked against the ADRs rather than against the issue state alone — the section says in its own words that an ADR declining to build something closes the decision, not the gap.

**The neighbouring pass the issue asks for, with its result.** Every `#N` in the file was resolved against the API, not read: #731 is the only closed one, #691 is open and correctly listed under Toward 1.0. Nothing else in the file references an issue.

**Known gaps is now empty and says so** instead of being dropped. Its definition is load-bearing: it is what keeps a declined decision from being filed as a non-goal, and the `deferred`-label paragraph refers to it.

**One sentence was added that is not a deletion.** The old entry ended by noting that context a *caller* hands the send stays unclassified — and that half is still true. ADR-164 says so explicitly where it declines to extend the ceiling to the task input, and ADR-144 gives the reason: a declaration with nowhere to live. Deleting the entry without saying that leaves the next reader to re-derive it and file it as a gap, which is the exact failure this issue is about. It is named as *not* belonging in the section, because no code closes it and no issue tracks it.

**Not fixed here, and filed separately:** nothing detects this drift. The file has now been corrected by hand three times — 2e2d5431, 90780db1 and this one — and 90780db1's own message names why (`"sixteen sibling merges changed that answer while it waited"`). A guard would have to ask GitHub for issue state, which is a real design question rather than a line to add to a docs PR, so it gets its own issue.

Tests: none. No PHP is touched. The four `ci:test:repo` checks were run directly and pass — the worktree has no `.Build`, so captainhook was not installed and the commit hook did not run them for me.

Closes #776

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -20,13 +20,14 @@ closes them. The [`deferred`](https://github.com/netresearch/t3x-nr-llm/issues?q
 label marks issues an ADR explicitly deferred; it spans this section and
 Toward 1.0 alike, so it defines neither.
 
-- **[#731](https://github.com/netresearch/t3x-nr-llm/issues/731) — a per-call injected snippet has no trust-zone ceiling.**
-  The sources a *record* declares are classified and gated: snippets and skills
-  (ADR-144), and the configuration's system prompt (ADR-155), judged against
-  the zone of the model that actually serves the call (ADR-149). Context a
-  *caller* hands the send is not, because it has no per-record home a
-  declaration could live on — the same argument that keeps task input
-  unclassified. ADR-155's own `Revisit when`.
+None open. #731 was the last, and ADR-164 with ADR-165 closed it in code: a
+run's forced sources now bind against the same trust ceiling as a
+configuration's own, at assembly and again on resume.
+
+The task input stays unclassified and does not belong here. ADR-144 declined it
+for want of a per-record home the declaration could live on, and ADR-164
+restates that reason where it declines to extend the ceiling to it — no code
+closes it, so no issue tracks it.
 
 ## Toward 1.0
 


CybotTM:
Review record — the Copilot review did not run. Its quota is monthly and account-wide and was exhausted before this PR was opened, so no bot review exists for this diff and re-requesting cannot produce one. Recording what was checked instead of merging on a silent gate: `decision=APPROVED` here comes from `github-actions`, not from a reviewer who read the change.

The diff is `ROADMAP.md` alone, 8 insertions and 7 deletions.

What I verified rather than assumed:

- **That the gap is closed by code, not by a decision.** The section states in its own words that an accepted ADR declining to build something closes the decision and not the gap. So the entry could only be removed if code closed it — ADR-164 binds a forced source against the ceiling and ADR-165 does it again on resume. Read both before editing.
- **That the neighbouring entries are current.** Every `#N` in the file resolved against the API: `#731` closed, `#691` open and correctly placed under Toward 1.0. Nothing else in the file references an issue.
- **That the half of the old entry which is still true survives.** Caller-supplied context stays unclassified, ADR-144 gives the reason, and ADR-164 restates it where it declines to extend the ceiling. Deleting the entry without saying so invites the next reader to re-file it as a gap, which is the failure #776 is about.
- **That the four repo checks pass.** Run directly, since this worktree has no `.Build` and captainhook was therefore not installed to run them on commit.

One correction to my own earlier claim in this thread: the first staleness scan ran against the `main/` worktree, which was ten commits behind `origin/main`. Re-run against `origin/main`; `ROADMAP.md` is byte-identical between the two, so the result stands, but it was luck rather than method.

No PHP is touched, no CHANGELOG entry is required, and `check-changelog-unreleased.php` agrees.

Merging on that basis.

PR #806, opened 2026-08-18
fix(build): accept the security commit type the hook already sees on main

### LINKED ISSUE #775: captainhook rejects the "security" commit type that CLAUDE.md documents as valid
`CLAUDE.md` documents the allowed conventional-commit types as `feat|fix|chore|docs|style|refactor|perf|test|build|ci|revert|security`, but the hook that enforces them does not accept `security`:

```
"regex": "#^(feat|fix|chore|docs|test|refactor|style|ci|perf|build|revert)(\\([^)]+\\))?!?: .{1,72}$#"
```

(`Build/captainhook.json`, the `commit-msg` Regex action.)

So a commit typed the way the contributor guidance says is valid gets rejected locally, and the natural next move — reaching for `--no-verify` — is the wrong lesson to teach.

Two ways to close it, and the choice is a real one: add `security` to the regex, or drop it from `CLAUDE.md` and let security fixes be typed `fix`. The second keeps the type list aligned with the changelog sections; the first makes security work greppable in the log. Either is defensible, but the two surfaces have to agree.

Found while preparing a security fix that would have been typed `security(context):`.

CybotTM:
AGENTS.md documents the allowed conventional-commit types as including `security`. The `commit-msg` Regex action in `Build/captainhook.json` did not list it, so a commit typed the way the contributor guidance says is valid was rejected locally — and the natural next move, `--no-verify`, is the wrong lesson to teach.

#775 framed this as a real choice: add `security` to the regex, or drop it from the docs and let security fixes be typed `fix`. The repository has already answered it, in three places that were checked rather than assumed.

- **32 commits use the type**, two of them on `main` — `5339098f` and `63a9e40a`, via #753 and #730.
- **Two merged PR titles are of that form**, so the shared `pr-quality` check accepts it. That surface is not this repository's to change, which settles the direction: aligning the docs down to the hook would have put them out of step with a workflow we do not own.
- **CHANGELOG.md carries five `### Security` sections.** The issue offered "keeps the type list aligned with the changelog sections" as the argument for dropping it, but Security *is* one of Keep a Changelog's sections — that argument points the other way once the file is read.

So the hook was the only surface that disagreed, and it was never right: `security` is absent from the file's first version (`ab381010`, the grumphp → captainhook migration), not removed from it later.

## Verification

A before/after on the regex, not a reading of it. `security(agent): bind the input to its turn` — the exact subject of `5339098f` — is rejected by the old pattern and accepted by the new one.

The controls matter more than the fix here, because widening an alternation is exactly how a guard stops guarding. All still refuse after the change:

| Subject | Expected |
|---|---|
| `securityfoo: not a type` | rejected — the alternation did not become a prefix match |
| `nonsense(x): whatever` | rejected |
| `fix: ` (empty subject) | rejected |
| `feat: ` + 73 characters | rejected |

Tests: no PHP is touched. The four `ci:test:repo` checks pass.

No CHANGELOG entry: this changes a contributor-side hook, not behaviour of the extension.

Closes #775

### DIFF
--- a/Build/captainhook.json
+++ b/Build/captainhook.json
@@ -11,7 +11,7 @@
             {
                 "action": "\\CaptainHook\\App\\Hook\\Message\\Action\\Regex",
                 "options": {
-                    "regex": "#^(feat|fix|chore|docs|test|refactor|style|ci|perf|build|revert)(\\([^)]+\\))?!?: .{1,72}$#"
+                    "regex": "#^(feat|fix|chore|docs|test|refactor|style|ci|perf|build|revert|security)(\\([^)]+\\))?!?: .{1,72}$#"
                 },
                 "conditions": []
             }


CybotTM:
Review record — the Copilot review did not run. Its quota is monthly and account-wide and was exhausted before this PR existed, so `decision=APPROVED` here comes from `github-actions` and not from a reviewer who read the diff. Recording what was checked instead of merging on that.

The diff is one line of `Build/captainhook.json`: `security` added to the `commit-msg` type alternation.

The direction was not a judgement call in the end — three surfaces already answered it, and each was read rather than assumed. 32 commits use the type, two of them on `main` (`5339098f`, `63a9e40a`); two merged PR titles of that form passed the shared `pr-quality` check, which is a workflow this repository does not own and therefore cannot be aligned downwards; and `CHANGELOG.md` carries five `### Security` sections, so Keep a Changelog's own section list agrees too. The hook was the only surface that disagreed, and it never agreed: `security` is absent from the file's first version, not removed from a later one.

The risk in this change is widening an alternation until it stops discriminating, so that is what the verification targets. Before: `security(agent): …` rejected. After: accepted. Still rejected after: `securityfoo:` (so the alternation did not become a prefix match), an unknown type, an empty subject, and a 73-character subject.

No PHP is touched. The four `ci:test:repo` checks pass.

Merging on that basis.

PR #807, opened 2026-08-18
test(adr): check the code citations in the ADR corpus against the tree

CybotTM:
The ADR corpus carries 74 citations of the form `File.php:NNN`, and nothing read any of them. `AdrLifecycleTest` checks form inside the corpus; `AdrReferenceIntegrityTest` checks that ADR filenames named from outside it resolve, and its docblock says outright that this is the part it does not cover — "a line number that has drifted still points at a line that exists".

## What the census found that #793 did not

21 of the 74 citations point into `.Build/vendor`: TYPO3 core (`DataHandler.php`, `RootLevelCapability.php`, `BackendUtility.php`, …), `cms-install`, and the sibling extension `nr-vault`. 18 of them are in ADR-169 alone. That code is not committed and its line numbers move with every patch release of a dependency this repository does not pin, so no test here can check them — and they are the most fragile citations in the corpus, not the least.

They are declared in a hand-maintained list instead. A new one fails the third assertion rather than being silently skipped by the resolver, which is the difference between a list and a blind spot.

The remaining 53 resolve to exactly one file each. Basename resolution is unambiguous today, and the resolver asserts that rather than picking the first match.

## Assertions, each seen to fail

A guard that has only been seen green proves it runs, not that it catches anything. Every branch was broken on purpose and the failure observed:

| Control | Result |
|---|---|
| citation to `NoSuchFileAnywhere.php:12` | `everyCitationNamesAFileThatStillExists` + the list assertion fail |
| citation to `ResumeCoordinator.php:204` (blank line) | `noCitationPointsPastTheEndOfItsFileOrAtABlankLine` fails |
| citation to `ResumeCoordinator.php:99999` (past EOF) | same assertion fails |
| citation to `DataHandler.php:1`, undeclared for that record | file + list assertions fail |
| a second `ResumeCoordinator.php` added to the tree | all three fail on the ambiguity assertion |

Each was reverted and the suite returned to green; `git status` was clean afterwards.

## What it does not catch

Stated in the docblock, not only here: **a line that moved onto different but non-blank code.** That is the common case and the dangerous one — the record then asserts something the cited code does not say, and the citation's precision is what makes a reviewer trust it. Nothing mechanical can tell the difference, which is the argument for citing a symbol rather than a line number wherever the prose allows.

## One correction to the issue

#793 gives ADR-171 citing `ResumeCoordinator.php:204` as confirmed rot. That is already repaired on `main` — the record cites `:205`, which is where `d3a8d718` moved the check. So this test does not fix a live defect. It stops the next one from being invisible, and the corpus is clean as of this commit.

I found that by re-running the census against `origin/main` rather than against the `main/` worktree, which was ten commits behind when I first measured.

#793 stays open: it also asks for the convention (symbol over line number) and covers the two counting issues, #791 and #792. This is the enforceable part.

Tests: `unit` (exit 0), `phpstan` level 10 clean, `cgl` and `rector -n` both stable at PHP 8.2 after applying.

Refs #793

### DIFF
--- a/Tests/Unit/AdrCodeCitationTest.php
+++ b/Tests/Unit/AdrCodeCitationTest.php
@@ -1,0 +1,272 @@
+<?php
+
+/*
+ * Copyright (c) 2025-2026 Netresearch DTT GmbH
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+declare(strict_types=1);
+
+namespace Netresearch\NrLlm\Tests\Unit;
+
+use PHPUnit\Framework\Attributes\CoversNothing;
+use PHPUnit\Framework\Attributes\Test;
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+use SplFileInfo;
+
+/**
+ * Checks the `File.php:NNN` citations in the ADR corpus against the tree.
+ *
+ * A line number is invalidated by any edit above it in the cited file, and
+ * nothing failed when that happened: ADR-171 cited ResumeCoordinator.php:204
+ * for an authorisation check that a later commit pushed to :205, and the record
+ * went on reading like a verified citation.
+ *
+ * What this catches, exactly:
+ *
+ * - a citation naming a file that is no longer in the tree, or whose basename
+ *   became ambiguous;
+ * - a citation pointing past the end of its file, or at a blank line.
+ *
+ * What it does NOT catch, stated so nobody reads more into it: a line that
+ * moved onto different but non-blank code. That is the common case and the
+ * dangerous one — the citation then asserts something the cited code does not
+ * say, and its precision is what makes a reviewer trust it. No test can tell
+ * the difference, which is the argument for citing a symbol rather than a line
+ * number wherever the prose allows it.
+ *
+ * Citations into TYPO3 core and sibling extensions cannot be checked here at
+ * all: that code lives in .Build/vendor, which is not committed, and its line
+ * numbers move with every patch release of a dependency this repository does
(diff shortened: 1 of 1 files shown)


CybotTM:
Review record — no Copilot review exists for this diff (monthly, account-wide quota exhausted), so the green gate is not a read of the change. What was checked:

**The test was built from a census, not from the issue.** #793 said 74 citations and one confirmed rot. Re-derived against `origin/main`: 74 is right, the rot is already repaired, and the 74 are **not one population** — 21 point into `.Build/vendor` (TYPO3 core, `cms-install`, `nr-vault`), 18 of them in ADR-169 alone. That code is not committed and its line numbers move with every patch release of an unpinned dependency, so no test here can check them. They are declared in a hand-maintained list so a new one fails rather than being skipped by the resolver.

**Each of the three assertions was seen to fail** — a missing file, a blank line, a past-EOF line, an undeclared vendor path, and an ambiguous basename each produce the failure they exist for. Reverted, green, `git status` clean.

**What it does not catch is in the docblock, not only in the PR:** a line that moved onto different but non-blank code. That is the common case and the dangerous one, and nothing mechanical distinguishes it.

**Cross-checked against the siblings in flight**, because this test reads the whole corpus and two other open PRs edit ADR text: none of #808, #810 or #811 adds a `File.php:NNN` citation, and none touches ADR-140 or ADR-169, the two records whose vendor citations this test hard-codes. So merge order among them is free.

`phpstan` level 10 clean, `unit` exit 0, `cgl` and `rector -n` stable at PHP 8.2 after applying.

Merging on that basis.

Issue #809, opened 2026-08-18
A forced source dropped before a run starts is silent


CybotTM:
When a run is queued with a forced snippet or skill and that record is switched off before the run starts, the source is dropped and nothing says so. The run then does not do what the person who queued it asked for, and the only way to find out is to compare the request against the transcript.

This has always been true for snippets — `AgentRunRequestCodec::snippetsByUids()` has gone through the active-only lookup since it was written. #781 asked for the two source kinds to stop disagreeing, and PR #810 does that by giving skills the same rule. So the inconsistency is gone and the silence is now uniform, which is why it gets its own issue rather than staying a clause in that one.

## Why this is not just logging

A log line answers "did it happen", which matters for support. It does not answer the question the operator actually has, which is "why did this run behave differently from the one I queued". That question is asked at the run, not at the log, so the answer has to live where the run is read.

The shape that already exists in this codebase is ADR-121's `dropped_turns`, recorded on the user line of a conversation whose context window overflowed. It is the same class of event: something the caller supplied did not reach the model, recorded next to the thing it happened to.

## What has to be decided

- **Where it is recorded.** On the run record, so the runs module can show it; or in the injected-context readout, next to the sources that *did* make it. The second is where a reader is already looking at the forced set.
- **Whether a dropped forced source should ever refuse the run instead.** For a queued run this is defensible in a way it is not for a resume: nothing has been sent yet, so refusing costs nothing but a message. ADR-165 deliberately does not refuse a resume over a missing source, and that reasoning does not carry over.
- **Whether "deleted" and "switched off" should read differently.** Both resolve to nothing today. They are different operator actions and a reader would want to tell them apart.

## Scope

Both source kinds and both composition paths — the playground's synchronous send and the queued run's dequeue. The resume path is out of scope: ADR-166 and ADR-175 keep a deactivated source resolving there on purpose, so nothing is dropped and there is nothing to report.

Split out of #781. Ref ADR-121, ADR-165, ADR-166.

PR #810, opened 2026-08-18
fix(agent): resolve a forced skill by the same rule as a forced snippet

### LINKED ISSUE #777: Forced skills are rebuilt twice with different ordering, changing which source a refusal names
Two places rebuild a run's forced skill set from persisted uids, and they order the result differently. On a classification tie that changes which source gets named.

`AgentRunRequestCodec::skillsByUids()` (`:216-237`) indexes `findAll()` by uid and then iterates the **persisted uid list**, so the result keeps the order the run was started with. `ToolLoopService::augmentationFrom()` (`:178-185`) iterates `findAll()` directly and appends matches, so the result keeps `SkillRepository::$defaultOrderings` — `name ASC` (`:23-25`).

Order is not cosmetic here. `InputContextClassification::withStricter()` ends in `$this->effective->isAtMost($other->effective) ? $other : $this` (`:84`), so on **equal** data class the later source in the fold wins — and it is that source's name that the refusal message and the governance row carry.

A run started with skills [B(CONFIDENTIAL), A(CONFIDENTIAL)] therefore blames B before it suspends and A after it resumes. Same ceiling, same outcome, different name in the audit and in the message the operator reads.

Two things worth doing together:

- one lookup instead of two copies, so the ordering rule has a single home;
- `SkillRepository` has no `findByUids()`, so both copies pay a full `findAll()` — the snippet side already has the targeted lookup.

Low severity as a security matter: the ceiling itself is identical either way. It matters for whether an audit trail can be read back consistently, which is the property the governance record is for.

Ref: ADR-165, ADR-144.

### LINKED ISSUE #781: A queued run drops a deactivated forced snippet but keeps a deactivated forced skill
`AgentRunRequestCodec` rebuilds a queued run's forced set from persisted uids and treats the two source kinds differently: `snippetsByUids()` goes through the active-only `PromptSnippetRepository::findByUids()`, while `skillsByUids()` filters `SkillRepository::findAll()` by uid alone and therefore ignores the enabled toggle entirely.

So for a run sitting in the queue between enqueue and dequeue:

- deactivate a forced **snippet** and it is gone when the run starts;
- deactivate a forced **skill** in the identical situation and it survives.

This is not the ceiling hole that #766 fixes. On the queued path the dropped snippet's text has not entered a transcript yet, so nothing classified goes on the wire unclassified — the run simply starts without that snippet. It is the same "a forced source silently disappears" shape one hop earlier, with a different consequence: a run does not do what the person who queued it asked for, and nothing says why.

Two defensible answers, and the point of filing is that it should be one of them deliberately:

- **Consistency toward active-only** — a source switched off should not enter a run that has not started yet, and the skill half is the one that is wrong.
- **Consistency toward existence** — what was forced at enqueue time is part of the request, and deactivating a shared record should not silently edit somebody's queued work.

Either way the two halves should stop disagreeing, and whichever is chosen should be visible to the operator rather than silent.

Found while reviewing #779; pre-existing, and that change neither causes nor worsens it.

Ref: ADR-165, ADR-102, #766.

CybotTM:
Two issues in the same three functions. #781 is a forced skill and a forced snippet answering the same question differently; #777 is the same forced skill set being ordered two ways.

## What the issues got wrong, and what the code says

#781 described the disagreement as living inside `AgentRunRequestCodec`. It is wider: **three** places rebuild a forced skill set from persisted uids, and the playground has the same disagreement as the codec.

`AgentRunRequestCodec::skillsByUids()` also carried a docblock saying the missing `enabled` filter was deliberate — "forcing a skill overrides its global toggle, the same semantics the playground's force-inject control has". That reads like an intended rule, so it was checked rather than taken: no record decides it, and the playground does not behave that way — `availableSkills()` lists enabled skills only, so a disabled skill can reach the forced set only from a stale form or a hand-built body. The sentence described neither an intent nor the thing it claimed to copy.

## The decision comes from ADR-166, not from me

ADR-166 already wrote the rule down twice. `findByUids()` is "the lookup for a prompt being assembled *now*, and a snippet an operator switched off must not enter one", and "a fresh run that forces it gets it through the active-only lookup like any other".

Its "**the skill half was already correct**" is true of the path it was judging — the resume — and says nothing about the two that compose. So the fix direction is not a coin toss between #781's two options; the record picks one, and this change applies it to the path ADR-166 never looked at. ADR-175 records that and ADR-166 gains the matching `:Amended:`.

| Path | Composing or re-gating | Lookup |
|---|---|---|
| `ToolPlaygroundController::resolveForcedSkills()` | composing now | `findByUids()` — enabled only |
| `AgentRunRequestCodec::skillsByUids()` | composing now (dequeue) | `findByUids()` — enabled only |
| `ToolLoopService::augmentationFrom()` | re-gating sent text | `findExistingByUids()` — existence |

## The ordering half (#777)

The resume path iterated `findAll()` and so returned `$defaultOrderings`, `name ASC`; the two composition paths returned the persisted uid order. `InputContextClassification::withStricter()` keeps the **later** source on an equal data class, and that source's name is what the refusal message and the governance row carry — so one run blamed one skill before it suspended and another after it resumed. Same ceiling, same outcome, different name in the audit. All three now use the caller's order.

## Verification

Ten functional cases against real rows. The fixture is its own and its names sort **against** its uids — `21` is "Zulu", `22` is "Alpha" — so an order assertion distinguishes the caller's order from `name ASC` instead of passing under both.

Each assertion was seen to fail before being trusted:

| Control | Result |
|---|---|
| drop the `enabled` clause from `orderedByUid()` | `findByUidsSkipsDisabledSkills` fails |
| return query order instead of the caller's | 3 tests fail, including both `findExisting…` order cases |

Restored, green, `git status` clean.

**One existing unit test broke, and that was a real signal rather than noise.** `ToolLoopServiceTest::resumeRebuildsTheForcedSkillsToo` stubbed `findAll()`. Its assertion — a disabled skill still resolves on resume — is exactly what must not regress, so the double was moved to the new lookup rather than the test being relaxed. It now also asserts *which* uids reach that lookup, so the seam is covered here while the filtering is proven against the database, where a stub cannot flatter it.

Gates: `phpstan` level 10 clean, `unit` 7137 pass, `fuzzy` pass, `functional -d sqlite` on the touched classes 38 pass, `cgl` and `rector -n` both stable at PHP 8.2 after applying.

## Deliberately not in this PR

A forced source that disappears before a run starts is **silent** — now uniformly, where before it was silent for snippets and absent for skills. #781 asked for it to be visible. That needs somewhere to show it and a decision on whether a queued run should refuse instead, so it is filed as #809 rather than guessed at here.

The ADR is in this PR rather than ahead of it because it amends a decision instead of taking one.

Closes #781, closes #777

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -203,6 +203,24 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   A row suspended before this change has neither key and resumes exactly as it
   did. A uid that no longer resolves — the snippet was deleted meanwhile —
   contributes nothing rather than stranding the run.
+- A forced skill now resolves enabled-only where it is being composed, and by
+  existence only where it is being re-gated — the rule ADR-166 already gave
+  snippets, applied to skills (ADR-175).
+
+  A run queued with a forced skill that an operator disabled before the run
+  started used to keep that skill, while a forced snippet deactivated in the
+  identical situation was dropped. The two source kinds now answer the same
+  question the same way on the playground's synchronous send and on a queued
+  run's dequeue: a source switched off does not enter a prompt being assembled
+  now.
+
+  A resume is unchanged and deliberately so — its text is already in the
+  transcript, so dropping the source there would lower the ceiling of content
+  still going out.
+
+  `SkillRepository` gains `findByUids()` and `findExistingByUids()`. It is
+  `@internal`, so no frozen surface moves.
+
 
 ### Fixed
 
@@ -256,6 +274,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   so the profile had to be found again by hand, one screen after reading its
   name. The readout has always honoured a `profile` query parameter; only the
   link never sent one.
+- The same forced skill set is no longer ordered two different ways.
+
+  Three places rebuilt it from persisted uids; two kept the order the run was
+  started with and the resume path returned `name ASC`. On an equal data class
+  the later source in the fold wins, so one run named one skill in a refusal
+  before it suspended and a different one after it resumed — same ceiling, same
+  outcome, different name in the message and in the governance row. All three
+  now use the caller's order (ADR-175).
+
+
 
 
 ## [0.29.1] - 2026-08-13
--- a/Classes/Controller/Backend/ToolPlaygroundController.php
+++ b/Classes/Controller/Backend/ToolPlaygroundController.php
@@ -983,25 +983,7 @@ final class ToolPlaygroundController extends ActionController implements LoggerA
      */
     private function resolveForcedSkills(array $uids): array
     {
-        if ($uids === []) {
-            return [];
-        }
-
-        $byUid = [];
-        foreach ($this->skillRepository->findAll() as $skill) {
-            if ($skill instanceof Skill && $skill->getUid() !== null) {
-                $byUid[$skill->getUid()] = $skill;
-            }
-        }
-
-        $skills = [];
-        foreach ($uids as $uid) {
-            if (isset($byUid[$uid])) {
-                $skills[] = $byUid[$uid];
-            }
-        }
-
-        return $skills;
+        return $this->skillRepository->findByUids($uids);
     }
 
     /**
--- a/Classes/Domain/Repository/SkillRepository.php
+++ b/Classes/Domain/Repository/SkillRepository.php
@@ -76,4 +76,97 @@ class SkillRepository extends Repository
 
         return $query->count();
     }
+
+    /**
+     * Find enabled skills by uid, preserving the input order.
+     *
+     * Unknown and disabled uids are silently skipped.
+     *
+     * This is the lookup for NEW composition — a skill an operator switched
+     * off must not enter a prompt that is being assembled now. Do not merge it
+     * with {@see self::findExistingByUids()}: the two answer different
+     * questions and the difference is deliberate (ADR-175, mirroring the
+     * snippet pair ADR-166 introduced).
+     *
+     * @param list<int> $uids
+     *
+     * @return list<Skill>
+     */
+    public function findByUids(array $uids): array
+    {
+        return $this->orderedByUid($uids, true);
+    }
+
+    /**
+     * Find skills by uid regardless of their enabled flag, preserving the
+     * input order.
+     *
+     * Same contract as {@see self::findByUids()} — unknown uids are silently
+     * skipped, a deleted record still resolves to nothing — except that a
+     * disabled skill still resolves.
+     *
+     * This is the lookup for text that is ALREADY in a transcript: a resumed
+     * agent run re-loads its forced sources to re-gate them (ADR-165), and
+     * disabling a skill mid-run must not silently drop the classification of
+     * text that is still on the wire (ADR-166). "Disabled" means "not for new
+     * composition", not "already-injected text loses its class".
+     *
+     * @param list<int> $uids
+     *
+     * @return list<Skill>
+     */
+    public function findExistingByUids(array $uids): array
+    {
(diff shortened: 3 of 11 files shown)


CybotTM:
Rebased onto `main` after #801 merged, now `c3f3aac0`. Seven siblings landed while this waited, so both conflicting files were **rebuilt, not merged** — "ours holds only my additions" stops being true from the second merge of a series, which is the failure this repository already carried to `main` once.

`Documentation/Adr/Index.rst` and `CHANGELOG.md` were each reset to their `main` version verbatim and this branch's content re-inserted: the toctree line in ascending position, the two changelog blocks under their existing `### Changed` and `### Fixed` headings.

Verified after rather than assumed:

- toctree reads `… Adr172, Adr173, Adr174, Adr175` and every `Adr*.rst` on disk is listed;
- three `### ` headings in `[Unreleased]`, 16 bullets where `main`'s 14 plus this branch's 2 is expected, each of the two blocks present exactly once, no bullet appearing twice;
- no real conflict markers — matched as `^<<<<<<< ` / `^>>>>>>> ` with the trailing space, since a bare `=======` is RST underlining;
- `check-changelog-unreleased.php` passes.

**The review record still applies, and that is checked rather than claimed:** the change set is `11 files changed, 414 insertions(+), 60 deletions(-)` across the same eleven paths as before the rebase, so nothing in the reviewed code moved.

**One thing worth naming, because it was a real risk and not just a formality:** #807 merged while this waited, and it added a test that reads the whole ADR corpus and asserts every `File.php:NNN` citation. ADR-175 is new text in that corpus. I checked before enqueueing #807 that this branch adds no such citation; the rebased unit run now confirms it against the merged tree rather than against my reading — 7200 tests, exit 0.

`phpstan` level 10 clean; the ten new functional cases pass alongside the two existing ones.

CybotTM:
Review record — no Copilot review exists for this diff (monthly, account-wide quota exhausted), so `decision=APPROVED` comes from `github-actions` and not from a reviewer who read the change. What was checked:

**The issue was wrong about where the defect lives, and the docblock was the trap.** #781 placed the disagreement inside `AgentRunRequestCodec`. There are three copies of the lookup and the playground has the same one. The codec then carried a docblock calling the missing `enabled` filter deliberate — "forcing a skill overrides its global toggle, the same semantics the playground's force-inject control has". That reads like a decided rule, so it was checked instead of believed: no record decides it, and the playground does not do it — `availableSkills()` lists enabled skills only.

**The direction is ADR-166's, not mine.** It writes the rule twice — `findByUids()` is "the lookup for a prompt being assembled *now*", and "a fresh run that forces it gets it through the active-only lookup like any other". Its "the skill half was already correct" is scoped to the resume path it was judging. A queued run has composed nothing at enqueue, so dequeuing it is that assembly. ADR-175 records this and ADR-166 gains the matching `:Amended:`.

**Both new assertions were seen to fail.** Dropping the `enabled` clause fails one test; returning query order instead of the caller's fails three. Restored, green, tree clean.

**An existing unit test broke and was treated as a signal, not noise.** It stubbed `findAll()`; its assertion — a disabled skill still resolves on resume — is exactly what must not regress. The double moved to the new lookup rather than the test being relaxed, and it now also asserts which uids reach it, so the seam is covered in the unit test while the filtering is proven against the database.

**Post-rebase:** change set byte-identical at `11 files changed, 414 insertions(+), 60 deletions(-)`; `unit` 7200 pass on the merged tree, which also exercises #807's corpus test against the new ADR-175; `phpstan` level 10 clean; `cgl` and `rector -n` stable at PHP 8.2.

**Not in this PR, deliberately:** a forced source dropped before a run starts is silent — now uniformly so. That is #809, split out because making it visible needs a place to show it and a decision on whether a queued run should refuse instead.

Merging on that basis.

PR #811, opened 2026-08-18
fix(usecase): carry the pack's recommended profile into the governance link

### LINKED ISSUE #778: The pack page links to the governance readout without pre-selecting the recommended profile
A use-case pack declares a recommended governance profile, and the pack detail page links to the governance readout so an operator can compare it. The link does not carry the profile.

`UseCasePackController.php:116` builds `governanceUrl` with `routeUrl()` given only a controller and an action, and `Show.html:233` uses it as-is. The readout therefore opens with no profile pre-selected, and the operator has to find the recommended one by hand — immediately after reading its name on the page they just left.

The pack knows which profile it recommends at that point, so passing it as a route parameter is the whole fix, plus whatever the readout needs to honour a pre-selection.

Small, but it is the last step of the one workflow packs exist to support: install, read the recommendation, compare it against what is configured.

CybotTM:
The pack detail page prints the governance profile a pack recommends, then says "Compare this posture with what is in force" and links the readout. The link carried no profile, so the operator landed on an unselected readout and had to find the recommendation again by hand — one screen after reading its name.

## One thing the issue expected turned out to be already done

#778 asked for the route parameter "plus whatever the readout needs to honour a pre-selection". The readout needs nothing: `LlmModuleController::governanceAction()` already reads `profile` off the query string, and its own profile buttons already link that way. Only the sender was missing.

So this is `routeUrl()` gaining an optional extra-parameter argument and one call site using it. `GovernanceProfile` is a backed enum, so the value is the same string the buttons use.

## The test dispatches the controller, which is new here

`UseCasePackRenderTest` hands `governanceUrl` to a view as a fixed string. That is right for what those tests check — the template's markup — and it means **none of them could ever have caught this**, because the value under test was supplied by the test.

So this one runs the real action and asserts on the rendered body. Nothing in the suite dispatched a backend module action before, and three things were needed to make it work; each is commented where it sits rather than left as incantation:

- the request must exist **before** the controller is resolved — Extbase's `ConfigurationManager` captures the ambient one, which the neighbouring DI test already works around;
- `BackendViewFactory` resolves template paths from the route's `packageName`, so the request needs a `route` attribute;
- `ModuleTemplate` translates its own chrome from `$GLOBALS['LANG']`.

**Seen to fail first.** With the argument removed the assertion fails on the rendered body; restored, the fifteen tests in the class pass.

Gates: `phpstan` level 10 clean, `unit` 7137 pass, `functional -d sqlite` on `UseCasePack*` 32 pass, `cgl` and `rector -n` stable at PHP 8.2 after applying.

Closes #778

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -194,6 +194,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   usage and limit only when it refuses, so a row could not tell "near the limit"
   from "nothing measured".
 
+- The governance link on a use-case pack page now carries the pack's
+  recommended profile.
+
+  The page names the recommendation and then invites the operator to compare it
+  against what is in force. The link went to the readout with nothing selected,
+  so the profile had to be found again by hand, one screen after reading its
+  name. The readout has always honoured a `profile` query parameter; only the
+  link never sent one.
+
+
 ## [0.29.1] - 2026-08-13
 
 ### Fixed
--- a/Classes/Controller/Backend/UseCasePackController.php
+++ b/Classes/Controller/Backend/UseCasePackController.php
@@ -114,7 +114,13 @@ final class UseCasePackController extends ActionController
             'pack' => $useCasePack,
             'plan' => $this->installer->plan($useCasePack),
             'wizardUrl' => $this->wizardUrl(),
-            'governanceUrl' => $this->routeUrl('nrllm_overview', 'Backend\\LlmModule', 'governance'),
+            // Carries the pack's recommended profile as the readout's `profile`
+            // query parameter, so the comparison the hint asks for is one click
+            // rather than a hunt through the profile buttons. The readout has
+            // always honoured it; only the link never sent it.
+            'governanceUrl' => $this->routeUrl('nrllm_overview', 'Backend\\LlmModule', 'governance', [
+                'profile' => $useCasePack->recommendedGovernanceProfile->value,
+            ]),
             'toolsUrl' => $this->routeUrl('nrllm_tools', 'Backend\\Tool', 'list'),
             // The Editor Action Center lives in the editor-facing module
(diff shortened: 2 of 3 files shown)


CybotTM:
Review record — no Copilot review exists for this diff (monthly, account-wide quota exhausted), so the green gate is not a read of the change. What was checked:

**One premise of the issue was wrong and checking it removed most of the work.** #778 asked for the route parameter "plus whatever the readout needs to honour a pre-selection". The readout needs nothing — `governanceAction()` already reads `profile` off the query string and its own buttons already link that way. Only the sender was missing, so this is one argument and an optional parameter on a private helper.

**The existing tests could not have caught this, and that is why a new kind of test is here.** `UseCasePackRenderTest` hands `governanceUrl` to a view as a fixed string; it can only prove the template prints what it is given. Asserting a URL the controller builds needs the controller to build it, so this dispatches the real action. Nothing in the suite did that before — it needs the request to exist before the controller is resolved, a route carrying `packageName`, and `$GLOBALS[LANG]`. Each is commented where it sits rather than left as incantation, because the next person to dispatch an action will hit all three.

**Seen to fail first:** with the argument removed the assertion fails on the rendered body; restored, the fifteen tests in the class pass.

**Merge safety checked rather than assumed:** this branch and #801 both add to `[Unreleased]`. Merged together locally on top of `main`, the result has three headings, no bullet twice, and 14 bullets where 12 + 1 + 1 is expected — a conflict-free auto-merge of this file is exactly the case this repository has been burned by, so the content was counted rather than the exit code trusted. #810 does conflict and is deliberately held back for a rebase.

`phpstan` level 10 clean, `unit` 7137 pass, `functional -d sqlite` on `UseCasePack*` 32 pass, `cgl` and `rector -n` stable at PHP 8.2.

Merging on that basis.

Issue #812, opened 2026-08-18
Reopen ADR-119: a management surface makes the backend section question due


CybotTM:
ADR-169 section 6 recommends reopening ADR-119. Accepting ADR-169 deliberately did **not** do that, and this issue is where it goes instead.

## Why it is not part of the accepting change

ADR-119 pre-settles four answers for exactly this trigger: the section is called "AI", the identifier is `netresearch_ai` and never a bare `ai`, the flat entries are grouped by subject first, and old routes keep working through kept identifiers plus `'aliases' => ['nrllm']`.

Applying those creates a new **top-level backend section shared with `nr_ai_search`, `nr_repurpose` and the cowriter**. That is a decision about three other products' backend UI, and it should not arrive as a side effect of retiring a permission grant in this repository. ADR-119's own status already reads `Accepted (deferred — the placement is not finally settled, see Revisit)`, so nothing is misrepresented by leaving it until the question is put properly.

## What has actually changed since ADR-119 was written

Two things, both verified rather than recalled:

- **The count.** ADR-119 described twelve submodules. Fourteen are parented to `nrllm` today, and a fifteenth nr_llm module — `nrllm_aitasks` — sits outside the tree under `parent => 'web'`. That correction landed in #808; the record now says fourteen and names `nrllm_aitasks` as explicitly not part of the number.
- **The constraint that put it there.** The module menu drops every top-level module whose own access check fails, so a child of the admin-only `nrllm` would be invisible to non-admins (ADR-131). **A management surface has the same constraint.** Without reopening ADR-119 it becomes the second flat `web`-parented module, and ADR-119's "twelve flat entries do not move as they are" arrives by accretion rather than by decision.

## Has the trigger fired?

Not yet, by ADR-131's own account. Its trigger is "the first cross-consumer editor surface — a personal usage-and-budget view, a personal run history, or a lead-editor view over a group". `nrllm_aitasks` ships an actor-scoped run viewport, but ADR-131 states that "'Own runs' for editors is mostly the approver's view" and that agent runs "are currently started from admin surfaces". A personal run history is what that surface will become, not what it is.

So the honest position is: the trigger has not fired, and the **next** surface fires it. This issue exists so that surface does not ship first and settle the question by accident.

## What closing it needs

A decision on whether the section is created, taken with the three sibling extensions in view rather than for nr_llm alone — and if yes, the `:Status:` and `:Amended:` edits on ADR-119 plus the four pre-settled answers applied.

[...]

PR #813, opened 2026-08-18
docs(adr): accept ADR-169 and retire the management grant reservation

### LINKED ISSUE #691: Management grant — needs a management surface first

### LINKED ISSUE #768: Management architecture ADR: rootLevel, grant scope, module placement, enforcement points
`#691` records that a `MANAGE` grant needs a management surface first. Auditing what that surface would actually require turned up a blocker that changes the shape of the work, and two places where the project's own record now contradicts itself. Both need deciding before any code.

**The blocker: a non-admin cannot edit a single `tx_nrllm_*` record today.** All nine TCA tables are `rootLevel => -1` and none sets `security.ignoreRootLevelRestriction` (zero hits repo-wide). TYPO3 grants pid-0 record access to a non-admin on exactly that flag (`DataHandler.php:7669-7674`), so no amount of `be_groups` `tables_modify` opens them. "Replace the admin gate with a management grant" therefore does not describe a buildable change — the grant alone yields nothing editable.

The inverse is also true and matters more: the write paths that already exist bypass DataHandler entirely. `TaskWizardController` and `SetupWizardController` write through Extbase repositories and `persistAll()`, so no record permission, no exclude-field and no rootLevel check runs on that path. Whatever a management module exposes, the grant would be the *only* control in front of it — every allowed action needs its own explicit check. `exclude => true` on `api_key` is FormEngine-only protection and buys nothing there; `endpoint_url` has no exclude flag at all.

**Contradiction 1 — the grant's name.** ADR-130 (`:84-88`), ADR-131 (`:67-69`) and the enum docblock (`BackendUserGrant.php:26`) all reserve `tasks_manage`. `#691` describes a general `MANAGE` grant over providers, models, configurations and tasks. That scope is much wider than the reserved name. Two names stand in the record and nothing picks a winner.

**Contradiction 2 — module placement.** ADR-119 decided "Administration, for now" and named the revisit trigger: the first cross-consumer editor surface. For that case it already pre-settles the answers — section "AI", identifier `netresearch_ai`, entries grouped by Setup / Authoring / Operation, old routes kept as aliases, explicitly not a dozen flat modules. A second flat `web`-parented module would be a decision against ADR-119 without saying so. A non-admin cross-consumer management surface looks like exactly the trigger ADR-119 named.

**Routing policy is not a manageable object.** Governance profiles have no apply path (ADR-140, and the code says so twice), and the policy mode is a per-call argument. A grant over it would gate nothing — the ADR-023/117 failure `#691` warns about. It is dropped from the scope.

**What this issue is for:** an ADR that decides, together, before code exists —

1. which records a non-admin may actually manage;
2. `tasks_manage` or a wider management grant, and what happens to the two ADRs that reserve the narrow name;
3. which fields stay admin-only (API keys, endpoint URL, vault config, extension configuration);
4. how `rootLevel => -1` is solved — opening the TCA for every module including the list module, or a purpose-built write path inside the new module;
5. FormEngine or a purpose-built management UI;
6. whether ADR-119 is reopened and the surface lands under `netresearch_ai`;
7. which concrete controllers and services are the enforcement points, since ADR-130 requires the grant to arrive with its consumer.

Blocks `#691`.

Ref: ADR-130, ADR-131, ADR-119, ADR-140, ADR-117, ADR-023.

CybotTM:
#768 asks for "an ADR that decides, together, before code exists" — seven questions, listed. **That ADR already exists.** ADR-169 is 442 lines, covers all seven, carries a recommendation for each, and sat at `Proposed`. What was missing was the decision, not the record. So this accepts it instead of writing a second one.

## The decisions

| § | Recommendation | Disposition |
|---|---|---|
| 1 | Records on a page inside a web mount (Option P), **not** opening `security.ignoreRootLevelRestriction` | accepted |
| 2 | Manageable: `tx_nrllm_task`, `tx_nrllm_promptsnippet`, `tx_nrllm_configuration` (last one minus governance and spend fields) | accepted |
| 3 | Both wizards move onto the DataHandler or stay admin-only | accepted |
| 4 | FormEngine plus `exclude => true` (Option C) | accepted |
| 5 | Neither grant name; close #691 as answered | accepted |
| 6 | Reopen ADR-119 | **not accepted here** → #812 |
| 7 | No grant case; preset import and pack install write as the acting user | accepted |

**Section 6 is the one exception and the record now says so.** Reopening ADR-119 creates a top-level backend section shared with `nr_ai_search`, `nr_repurpose` and the cowriter. That is a decision about three other products' backend UI and should not arrive as a side effect of retiring a permission grant here. ADR-169's Consequences bullet promised the accepting change *would* reopen it; leaving that standing would have made the record false the moment it was accepted, so it is corrected rather than ignored.

## What this retires

`tasks_manage` was reserved in three places. It is retired, and the reason is ADR-130's own rule rather than a change of mind: after sections 2 and 4 the manageable records are authorised by `tables_modify` and `non_exclude_fields` — TYPO3's own permissions on the same fields — so the grant would sit in front of nothing. ADR-130 named constraint 4, ADR-131's bullet and `BackendUserGrant`'s class docblock all record that now.

**The `:Amended:` pairing was verified by breaking it**, not by reading it: removing ADR-130's half makes `AdrLifecycleTest::amendingAndAmendedRecordsPointAtEachOther` name the missing side. Restored, green.

## One premise I checked before signing it

Section 1's Revisit list said it was read off 14.3.5 and that "the 13.4 leg of the matrix was not enumerated". The whole Option P conclusion rests on it, and this extension supports `^13.4 || ^14.3`, so accepting on one leg would have been signing for both.

13.4 answers identically: the same three `RootLevelCapability` constants with `TYPE_BOTH = -1`, the same skipped early return for `TYPE_BOTH` in `isTableAllowedForThisPage()`, the same `admin || shallIgnoreRootLevelRestriction()` at pid 0 with doktype-only above it, and the same split in `hasPageContextPermission()` between `VirtualRecord::RootPage` and the web mount plus the `perms_*` bitmask. The Revisit bullet now records the verification instead of the gap.

## Citation cleanup, and a limit it demonstrates

Fourteen `File:NNN` citations pointed into the four files this change edits — and one of them, ADR-169's own reference to ADR-130 constraint 4, was **already wrong before I touched anything**. They are replaced with references carrying no line numbers rather than renumbered, because renumbering only resets the clock.

Worth naming: `AdrCodeCitationTest`, merged three hours ago in #807, does not catch this. A citation whose target moved is still a non-blank line in an existing file. Its docblock says exactly that, and this is the first change to demonstrate it.

## Personas

Accepted on ADR-171's personas while they are still unvalidated — NEXT-152 has no answer yet. That is recorded in the new `Disposition` section rather than left implicit, and persona 7 (the management-grant holder) is named as the one this retires, so a later answer from Stefan has a stated place to land: section 5 is what it would reopen.

Gates: `phpstan` level 10 clean, `unit` 7200 pass (68 of them ADR suites), `cgl` clean, `rector -n` clean at PHP 8.2, the three repo checks pass. No PHP behaviour changes — the only code edit is a docblock.

Closes #768

### DIFF
--- a/Classes/Domain/Enum/BackendUserGrant.php
+++ b/Classes/Domain/Enum/BackendUserGrant.php
@@ -23,8 +23,14 @@ namespace Netresearch\NrLlm\Domain\Enum;
  *
  * Each case maps to exactly one enforcement point — there is no wildcard
  * grant, and a case is only added TOGETHER with its consumer (a grant
- * nothing reads is worse than none). `tasks_manage` therefore arrives with
- * the editing module, not here.
+ * nothing reads is worse than none).
+ *
+ * This docblock used to reserve `tasks_manage` for the editing module.
+ * ADR-169 retired that reservation: the records a management surface would
+ * own are authorised by `tables_modify` and `non_exclude_fields`, so the
+ * enforcement point it was waiting for is TYPO3's, not ours. Do not add the
+ * case back without naming an action those two permissions do not already
+ * cover.
  *
  * The values are deliberately underscore-separated, not colon-namespaced
  * like {@see ServiceAccountScope}: TYPO3 strips `:|,` from custom
--- a/Documentation/Adr/Adr130BackendUserGrants.rst
+++ b/Documentation/Adr/Adr130BackendUserGrants.rst
@@ -7,7 +7,9 @@ ADR-130: Capability grants for backend users
 :Status: Accepted (constraint 3 enumerated the approval surfaces and the
    enumeration was not exhaustive — see :ref:`ADR-131 <adr-131>`)
 :Date: 2026-08-06
-:Amended: 2026-08-06 by :ref:`ADR-131 <adr-131>`
+:Amended: 2026-08-06 by :ref:`ADR-131 <adr-131>`; 2026-08-18 by
+   :ref:`ADR-169 <adr-169>` (named constraint 4's reserved ``tasks_manage`` is
+   retired rather than fulfilled)
 
 Context
 =======
@@ -207,11 +209,25 @@ Named constraints, each verified against the code
    and the gate would be decorative. There is no four-eyes check on that
    path: :ref:`ADR-172 <adr-172>` refuses a self-*approval*, and
    supplying input is not one.
-4. **``tasks_manage`` does not exist yet.** The list/wizard actions have
-   no per-action gate to migrate (they are module-gated only), and the
-   trait's JSON 403 body is the wrong shape for HTML module actions. The
-   grant arrives together with its consumer in the editing-module
-   milestone.
+4. **``tasks_manage`` is retired, not pending** — amended, see
+   :ref:`ADR-169 <adr-169>`. This constraint reserved the name and said the
+   grant would arrive with the editing module.
+
+   It will not, and the reason is this record's own rule rather than a change
+   of mind: a case is added together with its enforcement point. ADR-169
+   section 2 settles which records a non-admin may manage — ``tx_nrllm_task``,
+   ``tx_nrllm_promptsnippet`` and ``tx_nrllm_configuration`` — and section 4
+   settles that they are managed through FormEngine with ``exclude => true``
+   on the governance and spend fields. Both halves are then authorised by
+   ``tables_modify`` and ``non_exclude_fields``, which are TYPO3's own
+   permissions on the same fields. A ``tasks_manage`` case would sit in front
+   of nothing.
+
+   The original observation still holds and is why the name was reserved
+   rather than built: the list and wizard actions carry no per-action gate to
+   migrate, and the trait's JSON 403 body is the wrong shape for an HTML
+   module action. What changed is that the gate they lack turns out not to be
+   ours to add.
 5. **The record picker stays admin-only.** ``TaskRecordsController``
    reads arbitrary table rows with only a housekeeping-prefix exclusion —
    no ``tables_select`` check, no denylist for ``be_users``/``sys_log``/
--- a/Documentation/Adr/Adr131EditorModule.rst
+++ b/Documentation/Adr/Adr131EditorModule.rst
@@ -8,6 +8,8 @@ ADR-131: The editor-facing module
 :Date: 2026-08-06
 :Amends: :ref:`ADR-130 <adr-130>` (constraint 3: this module is a third
    approval surface its enumeration did not have)
+:Amended: 2026-08-18 by :ref:`ADR-169 <adr-169>` (its "``tasks_manage`` still
+   does not exist" bullet said the grant was pending; it is retired)
 
 Context
 =======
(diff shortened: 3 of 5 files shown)


CybotTM:
Review record — no Copilot review exists for this diff (monthly, account-wide quota exhausted), so `decision=APPROVED` comes from `github-actions` and not from a reviewer. What was checked before signing an accepted decision:

**That I was not writing a second record.** #768 asks for an ADR deciding seven questions. ADR-169 already is that ADR — 442 lines, all seven, one recommendation each, `Proposed`. Found by reading ADR-171 persona 7, which names it. Had I gone straight to drafting, this PR would have been a duplicate of a record that was only missing a decision.

**That the central premise holds on both supported majors.** ADR-169 section 1 rests on `DataHandler` behaviour and its own Revisit list admitted it was read off 14.3.5 only, while the extension supports `^13.4 || ^14.3`. Accepting on one leg would have been signing for both. 13.4 was fetched and read: identical constants, identical skipped early return for `TYPE_BOTH`, identical `admin || shallIgnoreRootLevelRestriction()` at pid 0, identical split in `hasPageContextPermission()`. The Revisit bullet now records the verification instead of the gap.

**That accepting it would not make it false.** Its Consequences bullet said the accepting change reopens ADR-119. This change deliberately does not — that creates a top-level section shared with three sibling extensions — so the bullet is corrected and the deferral is recorded in a new `Disposition` section with #812 as its addressee. An accepted record that promises something the acceptance did not do is worse than a proposed one.

**That the amend pairing is real.** Removing ADR-130's `:Amended:` half makes `AdrLifecycleTest::amendingAndAmendedRecordsPointAtEachOther` name the missing side. Restored, green.

**That the personas it rests on are unvalidated, and where that lands.** NEXT-152 has no answer. Recorded in `Disposition` rather than left implicit: persona 7 is the one retired here, so section 5 is the decision a later answer would reopen.

**Citations.** Fourteen line-number references pointed into the four edited files and one — ADR-169's own reference to ADR-130 constraint 4 — was already wrong beforehand. Replaced with references carrying no line numbers rather than renumbered. Worth naming: `AdrCodeCitationTest` from #807 does not catch this class, because a shifted line is still non-blank in an existing file. Its docblock says so; this is the first change to demonstrate it.

Gates: `phpstan` level 10 clean, `unit` 7200 pass (68 ADR-suite tests among them), `cgl` clean, `rector -n` clean at PHP 8.2, three repo checks pass. The only code change is a docblock; no behaviour moves.

Merging on that basis.

PR #814, opened 2026-08-18
docs(adr): ADR-176, a per-call outcome kept apart from the approval gate

CybotTM:
The decision behind #772, before any code. No behaviour changes here.

## The problem

We cannot tell whether a single AI answer was good. There is no measurement per request.

That blocks one thing: sending simple requests to a cheaper model. ADR-156 allows it only if quality does not drop, and that is exactly what we cannot check.

A quality value does exist, but it does not help. It is per model, from a test set. Comparing two models, it is the same on both sides. So it can never show a difference — the citation in ADR-156's second criterion could never answer the question it was cited for.

## The decision

One outcome row per call, on `correlation_id` — the key ADR-174 already writes cost against, so quality and cost join without a second identity.

**Approval is not quality.** An approval can be withheld for governance or content reasons that say nothing about the model. No writer reads an approval state and the readout has no column for one.

**No person on the row.** No `be_user`, and no per-editor readout anywhere. The run record already carries who ran it, because budgets need it there. Erasure rides on the installation's `be_users` lifecycle — that is the framework's job and nr_llm does not build a second one. The join through the run still exists, and the record says so instead of resting on "we store no names".

## Two measurements, never averaged into one number

| | What it is | Where it works |
|---|---|---|
| **explicit** | the editor rates the result | everywhere |
| **observed** | what happened to the text afterwards | only where a tool wrote a record |

That split was read off the code, not assumed. `AiTaskController` renders the result and the editor copies it by hand — no write target, nothing to observe. `EditorActionController` writes through the DataHandler against a named record.

So `observed` is not a cheaper version of `explicit` that happens to be available everywhere. A figure averaging both would compare an observable population against an unobservable one, and a canary's arm difference would carry that asymmetry instead of the models' quality.

## Two things the draft turned up

**The observed signal is an event, not a time window.** The write tools refuse the draft workspace and write live; `CreateContentElementDraftTool` creates its element `hidden`. So "draft" here is a hidden live record, not a workspace version, and there is no publish or discard to measure. What the human does next is in `sys_history`: unhide, change first, or delete. A record still hidden gets no outcome — "undecided" is honest and is not a third quality value.

**And `observed` is blocked on something already deferred once.** Nothing persists *which* record a tool write produced: the uid sits in `ToolInvocation::$arguments` as free-form JSON, and `ToolEffectInterface` declares only `getEffect()` — a classification, not an identity. No write target, no join to `sys_history`.

That is ADR-122's territory. It declined the side-effecting tool contract because no tool had side effects, and its own status already records that premise as expired. Five builtins write through the DataHandler today. The prerequisite is not new work invented here; it is the part of ADR-122 whose reason for waiting is gone, and this outcome signal is the first reader it lacked.

## Sequencing

`explicit` ships first and alone — it needs no new contract. `observed` ships with the prerequisite above.

## Verification

The ADR suites caught a real mistake in this record. `AdrLifecycleTest` reads **every** `:ref:` inside an `:Amends:` field as an amend target, and my parenthetical linked ADR-060 — so the record claimed to amend it. Now only ADR-156 is named, and ADR-156 carries the matching `:Amended:`.

68 ADR-suite tests pass, `unit` 7200 pass, the three repo checks pass. No PHP is touched, so no CHANGELOG entry.

Refs #772

### DIFF
--- a/Documentation/Adr/Adr156PersistTheRoutingDecisionObserveComplexity.rst
+++ b/Documentation/Adr/Adr156PersistTheRoutingDecisionObserveComplexity.rst
@@ -7,7 +7,9 @@ ADR-156: Persist the routing decision, and observe complexity without routing
 :Status: Accepted
 :Date: 2026-08-11
 :Amends: :ref:`ADR-142 <adr-142>` (both of its deferrals are lifted, one of them only halfway)
-:Amended: 2026-08-14 by :ref:`ADR-174 <adr-174>`
+:Amended: 2026-08-14 by :ref:`ADR-174 <adr-174>`; 2026-08-18 by
+   :ref:`ADR-176 <adr-176>` (its second activation criterion cited ADR-060,
+   whose per-model value cannot distinguish two canary arms)
 :Authors: Netresearch DTT GmbH
 
 Context
--- a/Documentation/Adr/Adr176PerCallOutcomeSeparateFromApproval.rst
+++ b/Documentation/Adr/Adr176PerCallOutcomeSeparateFromApproval.rst
@@ -1,0 +1,144 @@
+.. include:: /Includes.rst.txt
+
+.. _adr-176:
+
+==============================================================================
+ADR-176: A per-call outcome, kept apart from the approval gate
+==============================================================================
+
+:Status: Accepted
+:Date: 2026-08-18
+:Amends: :ref:`ADR-156 <adr-156>` (its second activation criterion cited
+    ADR-060 as "the measured quality signal"; that per-model value cannot
+    answer it, and this record supplies the one that can)
+:Authors: Netresearch DTT GmbH
+
+Context
+=======
+
+:ref:`ADR-156 <adr-156>`'s second activation criterion asks whether quality
+degrades on a population served by a cheaper model, and points at
+:ref:`ADR-060 <adr-060>` for the measurement. ADR-060 is the golden-set layer:
+a value per **model**, graded out of request by a CLI command nothing
+schedules. In a canary both arms carry the same model-level value, so the
+criterion evaluates to "no difference" by construction, whatever the models
+did. The citation has never been able to answer the question it is cited for.
+
+Nothing else fills the gap. Governance events record blocks and never allows,
+so there is no denominator. There is no rating, no acceptance signal and no
+feedback of any kind, and retries are not counted. A promotion rule phrased as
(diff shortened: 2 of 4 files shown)


CybotTM:
Review record — no Copilot review exists for this diff (monthly, account-wide quota exhausted), so the green gate is not a read of the change.

**What this decides, in one line:** we cannot tell whether a single answer was good, ADR-156 needs exactly that to allow cheaper routing, and the value it cites is per model and therefore identical on both sides of a comparison.

**Checked rather than assumed, and it changed the shape of the work twice:**

The split between the two measurements is not a preference. `AiTaskController` renders a result the editor copies by hand — no write target, nothing to observe. `EditorActionController` writes through the DataHandler against a named record. So "observed" reaches one path, not both, and averaging the two would compare an observable population against an unobservable one.

The observed signal is an event rather than a time window, which I only found by reading the write tools: they refuse the draft workspace and create hidden **live** records. There is no publish or discard to measure; what follows is in `sys_history`.

And it is blocked on something already deferred once. Nothing persists *which* record a tool write produced — the uid sits in `ToolInvocation::$arguments` as free-form JSON and `ToolEffectInterface` declares only `getEffect()`. That is ADR-122's contract, whose stated premise (no tool has side effects) its own status already records as expired; five builtins write today.

**The ADR suites caught a real mistake in this record**, which is worth naming because it is the kind a reader would not: `AdrLifecycleTest` reads every `:ref:` inside an `:Amends:` field as an amend target, and my parenthetical linked ADR-060 — so the record claimed to amend a document it only cites. Now only ADR-156 is named, and ADR-156 carries the matching `:Amended:`.

**On the drift check that was red before the rebase:** it was not this PR. The org template moved `harden-runner` to v2.21.0 while `checks.yml` still pinned v2.20.1, so the check was red on `main` and every branch. I reproduced the comparison locally with the templates repo's own `drift_compare.py` — that one line was the entire reported difference — and #803 was the fix. After rebasing onto it the local run reports "No drift detected", which CI now confirms.

Gates: `unit` 7200 pass, 68 of them the ADR suites; the three repo checks pass. No PHP is touched, so no CHANGELOG entry and no behaviour moves.

Merging on that basis, and reading "mach das" for #772 as covering it: this repository's rule is that an ADR lands before the implementation it decides, so the mandate contains this merge rather than merely permitting it.

PR #818, opened 2026-08-18
feat: caller-source attribution on the options path (ADR-177)

### LINKED ISSUE #816: Telemetry: caller-source attribution channel for downstream consumers (nr-llm-compat)
The new compatibility layer ([t3x-nr-llm-compat](https://github.com/netresearch/t3x-nr-llm-compat)) routes third-party AI extensions' provider calls through nr-llm. Its design (ADR-001 there) wants every such request annotated with its origin — `source.extension` (e.g. `ai_seo_helper`), `source.operation` (e.g. `requestAi`) — so telemetry doubles as an AI inventory of the installation.

Today there is no channel for that: `CompletionServiceInterface::complete()`/`completeJson()` take no caller metadata, and the telemetry row (`tx_nrllm_telemetry`) has no caller-source column. `chatWithConfiguration()`'s `$metadata` array exists but nothing persists a source key from it.

Needed on the nr-llm side (public-surface change → ADR first, per repo convention):

- a way for a consumer to declare its identity per call (metadata key on the options/metadata path, or a scoped decorator of the feature services),
- persistence in the telemetry row (e.g. `source_extension`/`source_operation` columns) and visibility in the analytics/status surfaces.

Until this lands, intercepted requests appear in telemetry under the serving configuration without a caller tag — the compat layer already passes nothing rather than inventing a channel.

CybotTM:
Closes #816. Implements ADR-177 (#817).

## What

A downstream consumer can now tag any call with its identity, and the telemetry row records it:

```php
$options = (new ChatOptions())->withCallerSource('ai_seo_helper', 'requestAi');
```

- **`AbstractOptions::withCallerSource(string $extension, string $operation = '')`** plus the two getters — the same channel and wire-leak rule as the idempotency key: call metadata, never part of `toArray()`, so it cannot reach a provider.
- **`CallMetadataFactory::callerSource()`** maps it into the pipeline metadata as the fifth disjoint producer (keys published as `TelemetryMiddleware::METADATA_SOURCE_*` constants); all 13 option-carrying entry points in `LlmServiceManager` add the term to their existing `budget() + idempotency()` sums.
- **`TelemetryMiddleware`** persists it through `TelemetryRecord` into two new `tx_nrllm_telemetry` columns, `source_extension` / `source_operation` (`varchar(64)`, truncated rather than write-failing, `''` default). An unannotated call writes `''` — indistinguishable from a pre-feature row, so existing rows, queries and consumers are untouched.

Driving consumer: the nr-llm-compat bridge layer, which will tag every intercepted third-party request once its dependency floor reaches a release carrying this.

api-surface: +3 methods on `AbstractOptions`, additive (regenerated; `### Added` in the CHANGELOG entry).

Deliberately out of scope per ADR-177: analytics/dashboard aggregation by source, and authentication of the claimed identity — telemetry is an observability surface.

## Tests

- `CallMetadataFactoryCallerSourceTest` (mapping, empty-extension guard, operation default, cross-option-type via `VisionOptions`, key disjointness),
- `ChatOptionsTest`: wither immutability + the `toArray()` wire-leak exclusion,
- `TelemetryMiddlewareTest`: annotated metadata persisted, unannotated stays `''`,
- `LlmServiceManagerTest`: `chat()` plumbs / omits the metadata from the options (mirroring the budget-plumbing tests),
- functional `CallerSourceTelemetryTest`: both columns reach the database, `''` defaults hold, overlong identities truncate instead of failing the write.

Local gate: cgl, PHPStan (level 10), unit (7212+), fuzzy green; functional for the touched dirs (`Tests/Functional/Service/Telemetry`, `Tests/Functional/Provider/Middleware`) green on sqlite — the full functional matrix is CI's. The Rector gate could not run locally (worktree `.Build` resolved under PHP 8.5 vs the pinned 8.2 — the platform_check case `make gate` names); CI's pinned Rector job is the verdict for it. Mutation proof: nulling the factory's extension lookup turns 5 tests red.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [Unreleased]
 
 ### Added
+- **Telemetry rows can name their caller** (`#816`, ADR-177). A downstream consumer tags its calls with `->withCallerSource('my_extension', 'operation')` on any options object; the identity is persisted on the telemetry row as `source_extension` / `source_operation` and never sent to the provider. Unannotated calls keep writing `''`, so existing rows and consumers are untouched. Driving consumer: the nr-llm-compat bridge layer.
 - **A self-approved run says so** (`#785`). An approval granted by the backend user who started the run is now labelled as such — on the run timeline, and in the *Recent runs* table of the Agent Runs inbox, which gains an *Approval* column. An approval granted by anybody else is labelled too, so the two are told apart at a glance instead of by comparing two uid numbers.
 
   Derived at read time from values that were already stored, so it is right for runs recorded before this change. The readout compares the two uids in one place, `Domain\Enum\ApprovalAttribution`, and both surfaces render its answer; ADR-172's four-eyes gate keeps its own, stricter comparison in `AiActorContext::isInitiatorOf()`.
--- a/Classes/Provider/Middleware/TelemetryMiddleware.php
+++ b/Classes/Provider/Middleware/TelemetryMiddleware.php
@@ -77,6 +77,16 @@ use TYPO3\CMS\Core\Context\Exception\AspectNotFoundException;
 final readonly class TelemetryMiddleware implements ProviderMiddlewareInterface
 {
     /**
+     * Context-metadata key naming the calling software (ADR-177), set by
+     * {@see \Netresearch\NrLlm\Service\CallMetadataFactory::callerSource()}
+     * from the options object; persisted as `source_extension`.
+     */
+    public const METADATA_SOURCE_EXTENSION = 'sourceExtension';
+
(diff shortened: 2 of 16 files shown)


CybotTM:
Bot review was unavailable (Copilot errored twice on 3bc5923c — quota). Reviewed the diff manually instead, and the review found a real gap: the StreamingDispatcher is a SECOND telemetry write site (ADR-174's "the two write sites move together"), so an annotated streamed call would have lost its identity. Fixed in 18df78ec with a unit test on the streaming record plus the ''-default assertion on the existing success-path test. Also verified: no options class builds toArray() via get_object_vars (no wire-leak path), all 13 metadata-sum call sites carry the new term and none were missed, and TelemetryRecord's new parameters are appended so existing named-argument callers are untouched.

PR #820, opened 2026-08-18
test(api): freeze enum backing values in the api-surface snapshot

### LINKED ISSUE #815: Render backed-enum backing values in the api-surface snapshot
`Tests/Unit/Api/Support/ApiSurfaceRenderer.php` renders enum cases name-only (`'case ' . $case->getName()`). A backing-value change — `case Read = 'read'` → `= 'reveal'` — therefore leaves the snapshot byte-identical and sails through green, although for persisted or wire-visible enums the VALUE is exactly the frozen vocabulary the snapshot exists to guard. The enum-case constant path cannot catch it either: `isEnumCase()` cases are skipped in the constant loop, and `getValue()` on a case would render as `array`.

nr-vault ported the renderer for its own snapshot (netresearch/t3x-nr-vault#306) and fixed this there during review: backed cases now render as `case Read = "read"` via `ReflectionEnumBackedCase::getBackingValue()` + `json_encode` (deterministic; pure enums keep the bare form), with a fixture test proving a value change classifies as breaking — see netresearch/t3x-nr-vault#319. `ApiSurfaceDiff::memberKey()` already keys on `case <Name>`, so the change lands in `changed` with a was/now pair; no classifier change needed.

Mirroring it here would (a) close the same gap for nr-llm's `@api` enums and (b) restore the "identical rendering rules" property the two ports otherwise share. Cost: the one `declaredLines()` hunk, one fixture enum + two renderer tests (both can be copied from the nr-vault PR), and a snapshot regeneration whose diff is additive-only (every existing `case X` line gains ` = "..."`).

CybotTM:
Closes #815.

Backed cases rendered name-only, so a backing-value change under a stable case name — `case Read = 'read'` becoming `= 'reveal'` — left the snapshot byte-identical while the persisted vocabulary changed. The enum-case constant path could not catch it either: `isEnumCase()` rows are skipped and `getValue()` on a case renders as `array`.

Backed cases now render as `case Read = "read"` (via `ReflectionEnumBackedCase::getBackingValue()` + `json_encode`, so the rendering stays deterministic); pure enums keep the bare form. A fixture enum plus two renderer tests prove the rendering — including the engine-declared `cases()/from/tryFrom/name/value` members, which reflect as declared on the enum and have been matrix-stable in the snapshot since its first version — and that a value change classifies as **breaking** with the correct was/now pair. `ApiSurfaceDiff` already keys cases by name, so no classifier change was needed.

The committed snapshot is regenerated: exactly 108 case lines gain their value, no other line moves, and no API changed — the rendering rule did. One correction to the issue text: it predicted an additive-only diff, but `memberKey` matches cases by name, so the re-rendered lines land in `changed` — which is also why the regeneration belongs in this PR rather than surprising the next surface change.

Ported back from nr-vault, which caught the gap while porting this renderer (netresearch/t3x-nr-vault#319); the two snapshots use the same rendering rules again.

**Gates:** run per `make gate` — CHANGELOG check, cgl, PHPStan, unit (SUCCESS), fuzzy (86 tests OK), functional sqlite (1705 tests SUCCESS) all green locally. The Rector leg did **not** run locally: this worktree's `.Build` was resolved under PHP ≥ 8.4 and the 8.2-pinned leg dies at `platform_check` (the documented environment case) — CI's Rector job is the only verdict for it on this branch. No ADR: no `@api` signature changes; the change is to the test infrastructure that guards them, and the rendering rule is documented in the renderer's own docblock.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,17 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- **The api-surface snapshot freezes enum backing values** (`#815`). Backed
+  cases render as `case Read = "read"` instead of the bare name, because for
+  `@api` enums the values are the frozen vocabulary (persisted rows, wire
+  formats, CSV round-trips) and a value change under a stable case name
+  passed the snapshot byte-identical. A fixture enum plus two renderer tests
+  prove the rendering and that a value change classifies as breaking. The
+  committed snapshot is regenerated — 108 case lines gain their value, no
+  other line moves, and no API changed; the rendering rule did. Ported back
+  from nr-vault, which caught the gap in review
+  (netresearch/t3x-nr-vault#319).
+
 - **ADR-130's constraint 3 is amended by ADR-131** (`#787`). The record called
   `agent_approve` doubly unreachable for non-admins because it named two
   approval surfaces and both were admin-gated. Both still are: `nrllm_runs` is
--- a/Tests/Unit/Api/ApiSurfaceRendererTest.php
+++ b/Tests/Unit/Api/ApiSurfaceRendererTest.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace Netresearch\NrLlm\Tests\Unit\Api;
 
 use Netresearch\NrLlm\Tests\Unit\Api\Fixtures\AddedMethod;
+use Netresearch\NrLlm\Tests\Unit\Api\Fixtures\BackedFixtureEnum;
 use Netresearch\NrLlm\Tests\Unit\Api\Fixtures\InheritsForeignConstructor;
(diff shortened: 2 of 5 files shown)


CybotTM:
Review note: Copilot's review quota is exhausted for this month (account-wide), so no bot review will arrive. Fallback per policy: this diff is the faithful back-port of the twice-reviewed nr-vault change (netresearch/t3x-nr-vault#319, whose independent adversarial review produced the enum-value finding this implements), re-read critically before pushing; the snapshot diff was verified to touch exactly the 108 case lines and nothing else. CI's Rector job is the authoritative Rector verdict for this branch (local 8.2 leg cannot run — platform_check; see the PR body). No findings.

CybotTM:
Self-review: f6379ffd7880ac25719d02457359c3c5f066d098

PR #824, opened 2026-08-19
docs(spec): non-admin record management, specified against the accepted ADR-169

CybotTM:
## What this is

An implementation specification for non-admin record management, derived from **[ADR-169](https://github.com/netresearch/t3x-nr-llm/blob/main/Documentation/Adr/Adr169RecordManagementUsesTypo3Permissions.rst) as accepted on 2026-08-18** — sections 1, 2, 3, 4, 5 and 7 as written, section 6 explicitly not.

An earlier version of this PR asked for the Option P / Option R decision. That decision had already been taken: §1's recommendation is Option P, it was accepted with the rest, and [#768](https://github.com/netresearch/t3x-nr-llm/issues/768) and [#691](https://github.com/netresearch/t3x-nr-llm/issues/691) were closed `COMPLETED` the same day. The specification is rewritten against the accepted record and no longer requests anything.

## What the accepted sections require

| ADR-169 | Accepted | Consequence in the spec |
|---|---|---|
| §1 | **Option P** — records on a page inside a web mount | FR-005 keeps `security.ignoreRootLevelRestriction` closed on all nine tables |
| §2 | `tx_nrllm_task`, `tx_nrllm_promptsnippet`, `tx_nrllm_configuration` | FR-001 permits exactly those; FR-002 forbids widening without a separate security decision |
| §3 | Both wizards bypass the DataHandler | folded into FR-004 |
| §4 | **Option C** — FormEngine plus `exclude => true` | FR-003, and FR-004 carries §4's precondition: each wizard either moves onto the DataHandler or its surface stays admin-only |
| §5 | No nr_llm grant; `tables_modify` governs | no grant appears anywhere in the spec |
| §7 | Preset import and pack install move onto the DataHandler; the kill switch stays admin-only | out-of-scope list names `tx_nrllm_tool_state` and why it cannot be expressed |

**§6 stays out.** Module placement is [#812](https://github.com/netresearch/t3x-nr-llm/issues/812) with ADR-119 as its addressee. No requirement here depends on it.

Two requirements exist because Option P's accepted costs are real. **FR-007**: moving existing pid-0 records is a deliberate operator action, never an upgrade side effect. **FR-008**: page storage is a permission boundary and not a scope — nothing filters by pid at runtime, so a record created on any page is live installation-wide, and the documentation has to say that rather than implying a storage folder scopes anything.

## Renumbered, and why

`specs/002-per-call-outcome/` landed on `main` while this branch sat behind it, so both claimed `002`. This branch is rebased onto current `main` and the directory is now `specs/003-management-architecture/` — resolved against what exists rather than against the tree the branch was cut from.

## The preset install rides along

`.specify/presets/typo3` and `typo3-llm` are what shaped this document, so the artefact and the thing that produced it land together. Stack: `lean` (10) → [`typo3`](https://github.com/netresearch/spec-kit-typo3) (5) → [`typo3-llm`](https://github.com/netresearch/spec-kit-typo3-llm) (3).

**The TYPO3 layer earned its place** — version range, backward compatibility, the operator action, documentation and suite-per-requirement all produced content.

**The AI layer did not.** Provider capability, fallback, what leaves the instance, untrusted output, cost, non-determinism, drift: none of it has purchase on a backend permission change. The spec records that as a stated non-applicability rather than answering emptily. Picking this issue to exercise the third layer was my misjudgement; #770 or #772 would have done it.

## One gate this PR tripped, and the fix

Installing the presets brought their own repository-level `AGENTS.md` files into `.specify/presets/typo3/` and `.specify/presets/typo3-llm/`, and `check-agent-doc-symlinks.php` — added in [#796](https://github.com/netresearch/t3x-nr-llm/pull/796) — demands the `CLAUDE.md` / `GEMINI.md` pair beside every `AGENTS.md` it finds. It failed here with four missing aliases: my own gate, tripped by my own change.

`.specify` joins that script's `SKIP` list rather than `EXEMPT`, because the tree is vendored rather than written here. `specify preset add` copies an installed preset in whole, including that preset repository's `AGENTS.md`, which documents the preset and not any scope of this extension. The aliases genuinely cannot be there — Spec Kit's `safe_extract_archive` rejects an archive containing a symlink, so the presets ship without them on purpose — and anything placed there by hand is destroyed by the next `preset add`, which rewrites the directory.

Probed both ways: green on this tree, and still red when `Tests/CLAUDE.md` is removed, so the skip narrowed the rule without disarming it.

## Type of Change

- [x] Documentation update

## Checklist

- [ ] I have run `make gate` — **partially, and here is exactly which parts.** After resolving dependencies in this worktree: `cgl` reports nothing to fix across 1426 files, all four `ci:test:repo` scripts pass, and `AdrReferenceIntegrityTest` — the one unit test this diff can reach, since it scans `specs/` — passes with 1792 assertions. Not run: `phpstan` and `rector`, neither of which analyses `Build/Scripts`, and the functional suite, which no file here touches. Saying which rather than reporting a gate I did not run end to end.
- [x] I have added tests that prove my fix/feature works — the check fix was probed in both directions, and the four repo-check scripts are the assertions for it. The specification itself changes no behaviour.
- [x] I have updated the documentation accordingly — the specification is the documentation here.
- [ ] I have added a `CHANGELOG.md` entry — no user-facing change.
- [ ] If this changes the public surface, the ADR landed before this PR — no public surface change, and ADR-169 landed and was accepted before this rewrite.


CybotTM:
Copilot was unavailable for this PR — the review quota is exhausted for the month and account-wide, so re-requesting cannot change it before the reset. Reviewed by hand instead; below is what was checked and what the diff does not cover.

**Scope.** 28 files, +1067/-4. Twenty-five of those are the two presets, copied in whole by `specify preset add` — vendored, not authored here. The three files this repository owns are `specs/003-management-architecture/spec.md` and two four-word constant changes wrapped in docblocks.

**The two code changes are the same fix twice**, and both were tripped by this PR's own content rather than found by reading. Installing the presets brought their repository-level `AGENTS.md` files into `.specify/presets/*/`, which two independent gates then swept: `check-agent-doc-symlinks.php` demanded the alias pair beside each, and `AgentDocsCountConventionTest::theSweepFindsEveryAgentDoc` failed on a changed set across all eight unit cells. Both now skip `.specify`, each docblock naming the other so they stay in step.

**Why skip and not exempt, or not just extend the pinned list.** The tree is vendored: the aliases cannot exist there, because `safe_extract_archive` rejects an archive containing a symlink, and anything placed there by hand dies at the next `preset add`. Adding the two paths to the pinned list would apply a convention that governs *our* agent docs to a preset repository's own documentation, and would pin paths that the next install may rewrite.

**Verification.** `make gate` passed end to end on `a6ef6359`: cgl, PHPStan, unit, fuzzy, Rector pinned to 8.2, functional against SQLite, plus the CHANGELOG check. The symlink check was probed in both directions — green on this tree, still red with `Tests/CLAUDE.md` removed — so the skip narrowed each rule without disarming it.

**How the second gate was missed locally, since it matters more than the fix.** I first ran `--filter AdrReferenceIntegrityTest` on the reasoning that it was the one unit test this diff could reach. Two tests sweep the repository, not one. The filter cannot say that; the whole suite says it in a single run, which is what found it the second time.

**What this review does not establish.** The specification describes a surface nobody has built. Its requirements were derived from ADR-169 as accepted, and each `§` citation was read in the record on `main` rather than carried from the issue — but no code exercises them yet, so their testability is argued, not demonstrated. The vendored preset copies also drift from their upstream repositories the moment those change; that is inherent to how Spec Kit installs presets, and nothing here detects it.

Merging on that basis.

CybotTM:
Self-review: a6ef635921274bfdb4d53cf95075074d5b8d4417

PR #826, opened 2026-08-19
ci(roadmap): fail when the roadmap lists an issue that is already closed

### LINKED ISSUE #805: ROADMAP.md drifts against issue state and nothing detects it
ROADMAP.md asserts which issues are open. Nothing checks that assertion, and it has now been wrong three times.

- 2e2d5431 — #688 and #689 had closed; the file still carried both as open gaps.
- 90780db1 — #690, #692 and #718-#725 had closed while the PR waited. Its own message names the mechanism: `"This PR was held back to last on purpose: it asserts which gaps are open, and sixteen sibling merges changed that answer while it waited."`
- #776 / #804 — #731 had closed with ADR-164 and ADR-165.

Each time it was found by a person reading the file, and each time the fix was to re-derive the state by hand. The cost is not the edit; it is that a closed item in the file sends the next reader to re-derive a decision that has already shipped, which is what #776 was filed about.

## Why this is not simply a check to add

The state lives on GitHub, so the check needs the API. That collides with two rules this repository already holds.

`ci:test:repo` is the place a repo-specific check goes, and the same script runs as a pre-commit hook. A hook that needs network access either fails on a train or degrades to a pass — and a check that silently passes when it cannot answer is the guard-that-is-not-a-guard this repo has been burned by before. Making it skip offline means the only place it really runs is CI, which is defensible, but it should be a decision rather than a side effect of where the code was put.

`GITHUB_TOKEN` is available in Actions, so the CI half is straightforward. The local half is the open question.

## Options, none obviously right

1. **CI-only, and honest about it.** Add it to `ci:test:repo` guarded on `CI`, so it runs in the shared `repo-checks` job and does nothing locally. Cheap. The guard exists exactly where the drift lands (a PR), and nowhere else.
2. **A scheduled job instead of a PR check.** The drift is caused by *sibling merges*, not by the PR that touches the file — so a PR check on a branch that does not touch ROADMAP.md never runs, and the file rots between PRs. A weekly run that opens an issue when a referenced issue has closed matches the actual failure mode better.
3. **Remove the assertion instead of checking it.** Stop naming issue numbers in the file and link the two label queries instead. The file becomes a description of the sections plus a live query. This eliminates the drift rather than detecting it, at the cost of the per-item prose that explains *why* each gap is open — which is most of the file's value.

Option 2 addresses the observed cause; option 3 addresses it more completely and loses the most. Worth a decision before code.

## Scope note

Whatever is built must also cover the ADR prose, or it will be a partial guard sold as a whole one. Both earlier corrections had to fix an ADR in the same change — 2e2d5431 corrected ADR-139's "What this does not close", 90780db1 corrected the same record again — because a record's notes describe a gap that later closed. That is the same defect in a different file, and a ROADMAP-only check would leave it.

Found while closing #776.

CybotTM:
ROADMAP.md asserts which issues are open, nothing checked it, and the assertion has been wrong three times — 2e2d5431, 90780db1 and #804. Every one was found by a person reading the file.

## Why it runs on a schedule and not only on pull requests

The drift is produced by **sibling merges**, not by the pull request that touches ROADMAP.md. 90780db1's own commit message says it: *"This PR was held back to last on purpose: it asserts which gaps are open, and sixteen sibling merges changed that answer while it waited."*

So a pull-request check would sit on a branch that never touches the file and never run. That is the whole reason #805 offered three options rather than "add a check".

## It found real drift on its first run

**#691 was still listed as open work.** ADR-169 retired the management grant and #813 closed it as *answered*. The entry is removed here. Both roadmap sections are now empty and say so — including why the placement question (#812) is deliberately not listed: it is a decision shared with `nr_ai_search`, `nr_repurpose` and the cowriter, not work this extension can schedule alone.

## And a fault in itself

Scanning every `#123` flagged **#731**, which #804 mentions only as history — the sentence naming the ADRs that closed it. That is not a claim it is open, and a check that cannot tell an assertion from prose would have to be silenced within a week.

Scoping the scan to the file's own entry form — a bullet `- **[#N](url) — title.**` — is what separates them. That form is not invented for the check: the file already states "every item in the two sections below is an open GitHub issue".

## The constraint that shaped the implementation

`script-check.yml` says in its own header that it forwards no secrets and runs **without an authenticated `GH_TOKEN`**, so a `gh`-based check is not buildable against it. This repository is public, so the script reads the issues endpoint anonymously; the anonymous limit is 60/hour per IP.

That makes a third state real, and it is handled rather than ignored: rate-limited or unreachable exits **2**, a warning the caller lets pass. A shared runner IP can exhaust the quota through no fault of ours, and a check that failed on that would be switched off. **"Wrong" and "could not ask" must not look the same.**

## Controls, each observed

| Case | Expected | Result |
|---|---|---|
| entry naming a closed issue | fail | exit 1, names it |
| entry naming an open issue | pass | exit 0 |
| entry naming a non-existent number | fail | exit 1 |
| closed issue mentioned in prose | pass | exit 0 |

## One honest limit

**The check currently passes vacuously** — no entries remain after removing #691, so there is nothing to verify. It starts guarding with the first entry added. The controls above are what show it works in the meantime.

Repo rules checked: no local job (thin caller of the shared reusable, `check-workflow-ownership.php` passes), and `drift_compare.py` against the org template reports no drift.

Closes #805

### DIFF
--- a/.github/workflows/roadmap-check.yml
+++ b/.github/workflows/roadmap-check.yml
@@ -1,0 +1,37 @@
+name: Roadmap Issue State
+
+# Fails when ROADMAP.md lists an issue that is already closed.
+#
+# Runs on a SCHEDULE as well as on pull requests, and the schedule is the part
+# that matters: the drift is produced by sibling merges, not by the pull request
+# that touches ROADMAP.md, so a pull-request-only check would never see it. The
+# three corrections this exists to prevent (2e2d5431, 90780db1, #804) were all
+# caused that way.
+#
+# Thin caller of the shared script-check reusable, so the checkout pin and
+# harden-runner stay maintained centrally. That workflow forwards no secrets and
+# runs without an authenticated GH_TOKEN — the script therefore reads the public
+# issues endpoint anonymously, and reports "could not determine" (exit 2, a
+# warning that passes) rather than failing when the shared runner IP has
+# exhausted the 60/hour anonymous quota.
+
+on:
+  schedule:
+    # Weekly. The drift accrues over merges, not over hours.
+    - cron: '17 6 * * 1'
+  pull_request:
+    paths:
+      - 'ROADMAP.md'
+      - 'Build/Scripts/check-roadmap-issue-state.sh'
+      - '.github/workflows/roadmap-check.yml'
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  roadmap-check:
+    uses: netresearch/.github/.github/workflows/script-check.yml@main
+    permissions:
+      contents: read
+    with:
+      check-cmd: bash Build/Scripts/check-roadmap-issue-state.sh || [ $? -eq 2 ]
--- a/Build/Scripts/check-roadmap-issue-state.sh
+++ b/Build/Scripts/check-roadmap-issue-state.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env bash
+#
+# Fails when ROADMAP.md names an issue that is already closed.
+#
+# WHY THIS EXISTS. The file asserts which issues are open, and that assertion
+# has been wrong three times (2e2d5431, 90780db1, #804). The cause is not
+# carelessness: the drift is produced by SIBLING merges, not by the pull request
+# that touches ROADMAP.md — 90780db1's own message says "sixteen sibling merges
+# changed that answer while it waited". A check that only ran on pull requests
+# touching the file would therefore never see it, which is why the caller also
+# runs this on a schedule.
+#
+# WHY UNAUTHENTICATED. The shared script-check workflow deliberately runs with
+# no GH_TOKEN and forwards no secrets. This repository is public, so the issues
+# endpoint answers anonymously; the rate limit is 60/hour per IP.
+#
+# EXIT CODES. 0 = every referenced issue is open. 1 = at least one is closed.
+# 2 = the state could not be determined (rate limit, network). Two is a WARNING
+# the caller lets pass, because a shared runner IP can exhaust the anonymous
+# quota through no fault of ours — and a check that fails on that would be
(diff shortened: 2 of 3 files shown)


CybotTM:
Review record — the Copilot review did not run (monthly, account-wide quota), so `decision=APPROVED` here is `github-actions`, not a reader. What I checked:

**The design was forced by a constraint I verified rather than assumed.** #805 offered a scheduled `gh`-based job. `script-check.yml` states in its own header that it forwards no secrets and runs without an authenticated `GH_TOKEN`, so that is not buildable against the shared reusable — and this repository defines no local jobs. The script therefore reads the public issues endpoint anonymously, which I tested (HTTP 200, `#731 state=closed`, limit 60/hour per IP).

**That makes a third state real, and it is handled rather than ignored.** Rate-limited or unreachable exits 2 — a warning the caller passes. A shared runner IP can exhaust the anonymous quota through no fault of ours, and a check that failed on that would be switched off within a week. "Wrong" and "could not ask" must not look the same.

**The check found real drift on its first run and a fault in itself.** #691 was still listed as open work after ADR-169 retired the management grant — removed here. And scanning every `#123` flagged #731, which #804 mentions only as history. Scoping to the file's own entry form separates the assertion from the prose; that form is the file's own rule, not invented for the check.

**Four controls, each observed:** closed entry fails, open entry passes, non-existent number fails, prose mention does not fire.

**One honest limit, stated in the PR body too:** the check now passes vacuously — no entries remain. It starts guarding with the first one added, and the controls are what show it works meanwhile.

Repo rules: `check-workflow-ownership.php` passes (thin caller, no local steps), `drift_compare.py` against the org template reports no drift, `unit` 7215 pass. No PHP touched.

Merging on that basis.

CybotTM:
Self-review: 162e8b67eb85d21ecfd685cf984a128f3d5caefb

PR #827, opened 2026-08-19
docs(adr): ADR-178 caller source on the cost path

CybotTM:
Records the decision before the implementation PR, as the repo's ADR rule requires.

**Why a second record and not just UI work:** ADR-177 persisted the caller identity on the telemetry row and noted that analytics aggregation "can follow as UI work once rows carry the columns". Following it up showed that assumption to be wrong. Every analytics figure — KPI totals, provider and model breakdowns, per-user report — reads `tx_nrllm_service_usage`, the daily cost aggregate. Telemetry carries no cost and is purged on its own retention schedule, and the two tables share no key a per-day aggregate could be joined on. Attributing spend therefore needs the identity on the cost path too, which is a persistence decision: a new column, a wider daily aggregation key, and one additive `@api` parameter on `trackUsage()`.

Also decided here: only the extension key travels to the cost table (the operation stays a telemetry detail — it would multiply the aggregation key without answering the cost question), and unattributed usage gets an explicit label rather than being hidden.

ADR-177 is edited in the same change (`:Amended:` plus a pointer in the affected consequence), as the lifecycle in `Adr/Index.rst` requires. `AdrLifecycleTest` and `AdrReferenceIntegrityTest` pass.

### DIFF
--- a/Documentation/Adr/Adr177CallerSourceAttribution.rst
+++ b/Documentation/Adr/Adr177CallerSourceAttribution.rst
@@ -9,6 +9,7 @@ ADR-177: Caller-source attribution on the options path
 :Status: Accepted
 :Date: 2026-08-18
 :Authors: Netresearch DTT GmbH
+:Amended: 2026-08-19 by :ref:`ADR-178 <adr-178>`
 
 Context
 =======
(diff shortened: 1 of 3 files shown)


CybotTM:
Copilot review failed (quota); self-reviewed: docs-only — the new record plus the two-sided link into ADR-177 (`:Amended:` field and a pointer in the affected consequence) and the toctree entry. AdrLifecycleTest and AdrReferenceIntegrityTest pass in CI.

CybotTM:
Self-review: 4a03a46700bb4a7904c24efa4cfd3f0067a9fe50

PR #828, opened 2026-08-19
docs(adr): ADR-179, a dropped forced source is recorded on the run

CybotTM:
The decision behind #809, before any code. No behaviour changes here.

**The problem:** a run queued with a forced snippet or skill loses it if someone switches that record off before the run starts. The run proceeds without it and says nothing. The only way to notice is to compare the queued request against the transcript.

ADR-175 made this uniform rather than lopsided — before it, a snippet disappeared and a skill survived — which is what turns it into one question with one answer.

## Three decisions

**Recorded on the run, not only logged.** A log answers "did it happen", which serves support. The operator asks "why did this run differ from the one I queued", and that is asked at the run. It travels on `RunAugmentation` from the rehydration that drops it and is written through `RunTrace` — ADR-151's channel, whose steps `Backend/AgentRun/Show.html` already renders. No new surface is invented.

**The run is not refused.** Refusing a queued run is defensible on cost — nothing has been sent, so it costs only a message. Rejected on *who pays*: switching off a snippet is routine operator maintenance, and making it fail other people's queued work turns a safe action into one nobody dares take. A run that proceeds and says so is recoverable; a refused run is somebody else's incident.

**"Deleted" and "switched off" read differently.** Both resolve to nothing today and a single "dropped" would flatten them. They are different actions with different remedies — one is reversible, the other is not — and a reader who cannot tell them apart has to go looking.

## Deliberately out

- **The resume path.** ADR-166 and ADR-175 keep a deactivated source resolving there on purpose, because its text is already in the transcript. Nothing is dropped, so a report would imply otherwise.
- **Notifications.** The record is readable at the run; it does not push. Whether a dropped source deserves a notification is the same open question as every other pending decision here, and answering it for one case would settle it for all of them by accident.

## One correction I made while writing this

I had told the user the anchor was step metadata, citing `droppedTurns` in `ToolLoopService`. That line is a **log** call, not a step. Its own comment names the real one — the context step ADR-151 records through `RunTrace`. The decision above uses that channel; the sentence I said out loud first was wrong.

## Why this is its own PR

The implementation is roughly the size of #823: a field on `RunAugmentation`, the codec computing the difference, the trace write, the template, and tests. Splitting the decision out means it can be argued before that exists, and it gives an early checkpoint rather than a long silent stretch — which is the pattern that went wrong earlier today.

68 ADR-suite tests pass; the three repo checks pass. No PHP is touched.

Refs #809

### DIFF
--- a/Documentation/Adr/Adr175ForcedSkillsBindByTheSnippetRule.rst
+++ b/Documentation/Adr/Adr175ForcedSkillsBindByTheSnippetRule.rst
@@ -9,6 +9,8 @@ ADR-175: A forced skill binds by the same rule as a forced snippet
 :Amends: :ref:`ADR-166 <adr-166>` (whose "the skill half was already correct"
     held for the resume path it was judging and not for the two composition
     paths it did not look at)
+:Amended: 2026-08-19 by :ref:`ADR-179 <adr-179>` (it made the two kinds agree on
+    which sources enter a run and left the silence when one does not)
 :Authors: Netresearch DTT GmbH
 
 Context
--- a/Documentation/Adr/Adr179ADroppedForcedSourceIsRecordedOnTheRun.rst
+++ b/Documentation/Adr/Adr179ADroppedForcedSourceIsRecordedOnTheRun.rst
@@ -1,0 +1,102 @@
+.. include:: /Includes.rst.txt
+
+.. _adr-179:
+
+==============================================================================
+ADR-179: A forced source that is dropped is recorded on the run
+==============================================================================
+
+:Status: Accepted
+:Date: 2026-08-19
+:Amends: :ref:`ADR-175 <adr-175>` (which made the two source kinds agree on
+    WHICH sources enter a run, and left the silence when one does not)
+:Authors: Netresearch DTT GmbH
+
+Context
+=======
+
+A run queued with a forced snippet or skill resolves that set again when it is
+dequeued. :ref:`ADR-175 <adr-175>` settled that both kinds resolve enabled-only
+there, because dequeuing is composition and a source an operator switched off
+must not enter a prompt being assembled now.
+
+What it did not settle is what the operator is told. Today: nothing. The source
+is dropped, the run proceeds without it, and the only way to notice is to
+compare the queued request against the transcript. The run does not do what the
+person who queued it asked for, and nothing says why.
+
+The silence is now uniform rather than lopsided — before ADR-175 a snippet
+disappeared and a skill survived — which is what makes it a question with one
+answer instead of two.
+
+Decision
+========
+
+**Recorded on the run, not only in a log.** A log line answers "did it happen",
+which serves support. The question an operator actually has is "why did this
+run behave differently from the one I queued", and that is asked at the run.
+The record therefore travels on :php:`RunAugmentation` from the rehydration
+that drops it, and is written through :php:`RunTrace` — the same channel
+:ref:`ADR-151 <adr-151>` uses for context accounting, so it lands in the step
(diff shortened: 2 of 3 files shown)


CybotTM:
Review record — the Copilot review did not run (monthly, account-wide quota), so `decision=APPROVED` is `github-actions` and not a reader. What I checked:

**The anchor claim was wrong when I first made it, and I corrected it before writing the record.** I had said `droppedTurns` in `ToolLoopService` hangs off step metadata. It is a **log** call. Its own comment names the real channel — the context step ADR-151 records through `RunTrace`, whose steps `Backend/AgentRun/Show.html` renders. The decision uses that, and the PR body says so rather than quietly using the right one.

**Each of the three decisions carries its reason, and two of them turn on something other than cost.** Not refusing the run is decided on *who pays*: switching off a snippet is routine maintenance, and making it fail other people's queued work turns a safe action into one nobody dares take. Splitting "deleted" from "switched off" is decided on *remedy*: one is reversible, the other is not, so a single "dropped" would send a reader looking.

**The scope exclusions were checked against the records they cite**, not asserted: ADR-166 and ADR-175 do keep a deactivated source resolving on a resume, so nothing is dropped there and a report would imply otherwise.

**`:Amends:` pairing:** ADR-175 carries the matching `:Amended:`. The pairing rule is enforced by `AdrLifecycleTest`, which I have seen fail on a missing half earlier today, so this is a checked property rather than a convention I followed.

68 ADR-suite tests pass; the three repo checks pass. No PHP is touched, so no CHANGELOG entry and no behaviour moves.

The implementation follows as its own PR — deliberately split, so the decision can be argued before the code exists.

Merging on that basis.

CybotTM:
Self-review: a1a247945849f9b6a1c8a412b40d62ac8559d16c

CybotTM:
Renumbered from ADR-178 to **ADR-179**, now `367ed088`.

#827 claimed 178 at 14:04, eleven minutes before this branch existed. I had checked the highest number on `main` — where 177 was the top — instead of across the open branches, which is the collision the worktree rule exists to prevent. The earlier claim keeps the number.

Renamed the record, its `.. _adr-179:` label and title, the toctree entry, and ADR-175's `:Amended:` reference. The ADR suites pass (68 tests), including the pairing check that would have caught a half-renamed reference.

The previous self-review attestation died with this push, as designed. It rested on the old head and this is a different one.

The toctree now reads … 177, 179 with a gap where #827's 178 lands. Whichever of the two merges second will conflict on that line and the resolution is to keep both in ascending order — mechanical, and worth naming here so it is not re-derived under time pressure.

CybotTM:
Review record for head `367ed088` — the previous attestation died with the renumbering push, as designed, so this is a fresh read of the current head rather than a re-post.

What changed since the last record is the number and every reference to it: the record file, its `.. _adr-179:` label, the title, the toctree entry and ADR-175's `:Amended:` back-reference. The ADR suites pass (68 tests), and the pairing assertion among them is what would fail on a half-renamed reference — I have watched it fail on exactly that earlier today, so it is a checked property here and not a convention I trusted.

The decision content is unchanged and reviewed as before: recorded on the run rather than only logged, the run is not refused, and "deleted" reads differently from "switched off". The two exclusions were checked against the records they cite — ADR-166 and ADR-175 do keep a deactivated source resolving on a resume, so nothing is dropped there.

74 checks pass, 0 fail, 0 unresolved threads. No PHP is touched, so no CHANGELOG entry and no behaviour moves.

Merging on that basis.

CybotTM:
Self-review: 367ed0880c68a0e2d11ba5849f12fa6737b25751

PR #829, opened 2026-08-19
feat(analytics): break usage and cost down by calling extension

CybotTM:
Makes the caller attribution from ADR-177 answer the question it exists for: **which extension spends what**. Stacked on #827 (the ADR); the diff against that branch is the implementation.

**Why the telemetry column was not enough.** Cost is aggregated daily in `tx_nrllm_service_usage`, and every analytics figure reads that table; telemetry carries no cost, is purged on its own retention schedule, and shares no key a per-day aggregate could join on. So the identity travels the cost path too: `source_extension` becomes a column on the usage table and part of its daily aggregation key — two extensions on one model on one day stay two rows. `UsageMiddleware` reads it from the same request metadata `TelemetryMiddleware` reads, so the cost row and the telemetry row cannot disagree about who called.

**In the backend:** the Analytics module gains a *By extension* chart next to provider/model/service and a per-extension table with cost, requests and tokens. Unannotated usage — wizard tasks, scheduler runs, anything that names no caller — is labelled *Unattributed* instead of being hidden or folded into another bucket. EN and DE labels included; the admin docs page gained a section stating what the numbers mean and that attribution is an unverified claim, not an access control.

**Surface:** `UsageTrackerServiceInterface::trackUsage()` grows one optional trailing `$sourceExtension`; `api-surface.txt` is regenerated and the diff is that one additive line. Rows written before this change carry `''` and stay unattributed — no migration invents an origin.

**Tests** (both guards were watched failing before they were kept):
- Aggregation key: removing `source_extension` from the tracker's lookup makes the two new functional tests fail with `1 is identical to 2` — the rows collapse into one.
- Middleware wiring: hard-coding the read to `''` fails the new unit test.
- Also new: unattributed-vs-attributed separation, varchar(64) truncation, and the analytics breakdown ordering with its `unattributed` label.

Gates run locally: cgl, PHPStan level 10, full unit suite (7217 tests), functional `-d sqlite` for the touched classes (103 tests). Rector was not run locally (its `-p 8.2` pin needs a second `.Build`); CI is the verdict there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [Unreleased]
 
 ### Added
+- **Analytics answers "which extension spent what"** (ADR-178). The usage
+  table carries the caller's extension key next to the money, so the
+  Analytics module gains a *By extension* chart and a per-extension table
+  with cost, requests and tokens. Calls that name no caller — wizard
+  tasks, scheduler runs, anything unannotated — are listed as
+  *Unattributed*. `UsageTrackerServiceInterface::trackUsage()` grows one
+  optional trailing `$sourceExtension` parameter (additive); rows written
+  before this change stay unattributed.
+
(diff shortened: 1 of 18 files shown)


CybotTM:
Self-review: b82dc0cbe0fd97dd21697c2112c4b1feb9fd014a

PR #839, opened 2026-08-20
ci: test MariaDB on the current LTS series, 11.8 and 12.3

CybotTM:
CI runs the MariaDB series that MariaDB still supports. Per the vendor's [maintenance policy](https://mariadb.org/about/#maintenance-policy) those are 10.11 (until 2028-02), 11.4 (2029-05), 11.8 (2028-06) and 12.3 (2029-06). 12.0 to 12.2 are rolling releases, not LTS — and Docker Hub last rebuilt `mariadb:12.2` on 2026-05-30 while 12.3 is rebuilt daily.

| Surface | Before | After |
|---|---|---|
| `ci.yml`, functional leg | `mariadb:10.11` | `mariadb:11.8` |
| `e2e.yml` | `mariadb:12.2` (rolling, stale image) | `mariadb:12.3` |
| `runTests.sh`, allowed `-i` | `10.5\|10.6\|10.11\|11.0\|11.4` | `10.11\|11.4\|11.8\|12.3`, default `11.8` |

The two legs run different series on purpose, so CI covers both current LTS lines at no extra cost.

## Why the pin existed, and why it can go

The functional leg sat on 10.11 because the shared workflow health-checked the DB service with a hardcoded `mysqladmin ping --silent`, which mariadb images from 11.0 no longer ship — the service never turned healthy and the job died with "Failed to initialize container" (#334, worked around in #337). That is fixed upstream ([typo3-ci-workflows#128](https://github.com/netresearch/typo3-ci-workflows/issues/128), landed in [#174](https://github.com/netresearch/typo3-ci-workflows/pull/174)), and this repository tracks that workflow `@main`.

## Old versions are mapped, not refused

Dropping `10.5`, `10.6` and `11.0` from the accepted set would have broken every existing `-i 10.5` in a Makefile, a README or someone's shell history. Such a version is mapped onto the LTS of its own series and the substitution is printed:

| Requested | Runs | |
|---|---|---|
| `10.5`, `10.6` | `10.11` | with a warning |
| `11.0`–`11.3` | `11.4` | with a warning |
| `11.5`–`11.7` | `11.8` | with a warning |
| `12.0`–`12.2` | `12.3` | with a warning |
| `10.11`, `11.4`, `11.8`, `12.3` | as requested | silent |
| anything else | — | `Invalid combination`, naming the supported set |

The warning is not decoration: a suite that silently ran a different engine than it was asked for is worse than a call that failed. `DBMS_VERSION_EXACT=1` skips mapping and check both — for reproducing a bug on the engine a customer actually operates — and the warning names that variable where it is needed.

## Why `runTests.sh` changes too

Its list offered neither 11.8 nor 12.3, so the versions CI runs could not be reproduced locally — the script was one version behind the fleet, not the other way round. 10.5, 10.6 and 11.0 leave the accepted set but keep working through the mapping above.

## Verified

- `./Build/Scripts/runTests.sh -s functional -d mariadb -i 11.8 -p 8.4 Tests/Functional/Schema/` → `SUCCESS`, 1 test, 9 assertions.
- `-i 12.2` end to end: warns, runs 12.3, suite green — so both the mapping and 12.3 as an engine are proven, not argued.
- Every row of the mapping table was exercised against the function as it stands in the file: the eight out-of-support versions each resolve to the LTS shown and print the warning, the four supported ones pass through silently, `9.9` fails and lists the supported set, and `DBMS_VERSION_EXACT=1 -i 10.5` keeps 10.5 and says nothing. Checked by invoking, not by reading the regex.
- `mariadb:11.8` and `mariadb:12.3` both reach `healthy` under the fixed health command; `max_connections` reads 151 on both, as on 10.11, so the note above `db-image` explaining why four parallel shards are safe still holds.
- TYPO3 sets no upper bound: `EXT:install` declares a minimum of 10.4.3 for MariaDB and no maximum (`DatabaseCheck/Platform/MySql.php:52`). The `<= 10.99.99` shown on get.typo3.org is release metadata, not an engine check.
- `shellcheck --severity=warning` on `runTests.sh` reports the same 21 pre-existing findings as `origin/main` — none new.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01GSptxPLHWsttu9FuqVkvYZ)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,8 +99,11 @@ jobs:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
   # MariaDB functional leg (one cell): keeps the MySQL-only branches —
   # ke_search MATCH...AGAINST retrieval — permanently exercised in CI.
-  # mariadb:10.11 LTS: images >= 11 ship only mariadb-admin, which fails
-  # the reusable workflow's hardcoded mysqladmin health check.
+  # mariadb:11.8 — one of the two current LTS series (11.8, 12.3); the e2e
+  # leg runs the other, so CI covers both. This sat on 10.11 because images
+  # >= 11 ship only mariadb-admin while the reusable workflow health-checked
+  # with a hardcoded mysqladmin, so the service never turned healthy
+  # (netresearch/typo3-ci-workflows#128, fixed there in #174).
   # The full matrix above stays on SQLite; this narrow second call exists ONLY
   # for the MariaDB functional leg.
   ci-functional-mariadb:
@@ -112,7 +115,7 @@ jobs:
       typo3-versions: '["^14.3"]'
       run-functional-tests: true
       functional-test-db: mariadb
-      db-image: 'mariadb:10.11'
+      db-image: 'mariadb:11.8'
       # Shard by test class here too. With the SQLite cells down to ~2.5min this
       # leg became the single longest job in CI at 6min, i.e. the whole
       # critical-path term of the wall clock.
@@ -122,7 +125,7 @@ jobs:
       # identifier — $originalDatabaseName . '_ft' . substr(sha1(static::class), 0, 7)
       # (FunctionalTestCase.php:361) — so each process creates its OWN database
       # (typo3_test_ft<hash>). No CREATE race, no shared state, and 4 processes
-      # are nowhere near mariadb:10.11's default max_connections of 151.
+      # are nowhere near mariadb:11.8's default max_connections of 151.
       #
       # No event split: this leg never uploaded coverage (upload-coverage is
       # unset here, so the reusable resolves coverage: none). It exists purely to
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -28,7 +28,9 @@ jobs:
       contents: read
     with:
       php-version: '8.5'
-      db-image: 'mariadb:12.2'
+      # 12.3, not 12.2: 12.0-12.2 are rolling releases, 12.3 is the LTS of
+      # the 12 series (and the only 12.x still built — 12.2 stopped in May).
+      db-image: 'mariadb:12.3'
       # Seed WAITING_FOR_APPROVAL/INPUT runs before Playwright so the a11y
       # suite can exercise the approve/deny + schema-input forms (ADR-109).
       # Default-mode override: the workflow sets up TYPO3 and starts the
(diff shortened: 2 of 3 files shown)


CybotTM:
**Review note.** The `copilot_code_review` ruleset could not deliver a review — the account's monthly Copilot quota is exhausted, so the bot returns an error in place of a review and re-requesting cannot change that before the reset. Reviewed by hand; recording what was checked.

- Version lifecycles taken from the vendor, not from memory: 10.11, 11.4, 11.8 and 12.3 are the series still in community support; 12.0–12.2 are rolling releases. Docker Hub's build dates corroborate it — 12.2 last rebuilt 2026-05-30, 11.8 and 12.3 daily.
- The first version of this branch simply removed `10.5|10.6|11.0` from the accepted set. That was a defect: every consumer with `-i 10.5` in a Makefile would have gained an `Invalid combination`. The second commit maps instead, and announces the substitution.
- Mapping verified per row against the function as it stands in the file, including the two cases that are easy to get wrong: a supported version must pass through **silently** (checked: zero messages for 10.11, 11.4, 11.8, 12.3), and `DBMS_VERSION_EXACT=1` must suppress both the mapping and the check (checked: `-i 10.5` stays 10.5, no output).
- End to end on the real script: `-i 12.2` prints the warning, runs 12.3, suite green. That is one run proving three things — the mapping fires, 12.3 works as a functional engine, and the script is not broken by the edit.
- `max_connections` read back as 151 from both 11.8 and 12.3, which is what the comment above `db-image` relies on when it argues four parallel shards are safe.
- TYPO3 sets no upper bound — `EXT:install` declares a minimum of 10.4.3 and no maximum (`DatabaseCheck/Platform/MySql.php:52`). The `<= 10.99.99` on get.typo3.org is release metadata, not an engine check.
- shellcheck: 131 findings, the same set as `origin/main`, none new. This repository's CI does not shellcheck `runTests.sh` at all (`ci:test:repo` covers agent-docs, badges, changelog, workflows), so nothing regresses either way.
- CI on this head: 73 pass, 0 fail. The two decisive cells are `ci-functional-mariadb / Functional Tests (8.4, ^14.3)` on 11.8 and `e2e / E2E` on 12.3.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01GSptxPLHWsttu9FuqVkvYZ)_

CybotTM:
Self-review: c7144478db2776f90edcc533836b0a351bbb3823

PR #842, opened 2026-08-20
feat(agent): say which forced source a queued run did not get

### LINKED ISSUE #809: A forced source dropped before a run starts is silent

CybotTM:
Closes #809, implementing ADR-179 (#828).

**The problem:** switch a snippet or skill off between enqueue and start and the queued run loses it — that is ADR-175's rule and it stays. Nothing said so. The run did something other than what was asked for, and the only way to notice was to diff the queued request against the transcript.

## What it does

The dequeue reports what it asked for and did not get, and **why**: `deactivated` for a record that is switched off, `gone` for one that no longer resolves.

Telling those apart is the whole decision — one is reversible, the other is not — and it needs the **second lookup**: a uid the enabled-only lookup skipped but the existence lookup finds is switched off. Without both, the two reasons collapse into one and the reader is told "dropped" without being told what to do about it.

It travels on `RunAugmentation`, is written through `RunTrace` as its own step kind before assembly, and reaches the run view through `RunStep`'s payload and the timeline's allow-list.

**Recorded on the resolve side, not inferred from the assembled messages** — a source that never arrived leaves no mark in them, and that absence is precisely what an operator cannot see today.

**Flattened to `kind#uid reason`, not counted.** `snippet#41 deactivated` says what to do; `2 dropped` flattens the distinction the record exists to make.

A run that dropped nothing records **no step**, so the step's presence is the signal — an empty step on every run is noise, and then nobody looks. The resume path is untouched: ADR-166 and ADR-175 keep a deactivated source resolving there, so nothing is dropped and a step would imply otherwise.

## An existing test caught a real defect

`ToolLoopServiceAssemblyOrderTest` pins that the enabled-only lookup runs **once** per dequeue. My first version called it a second time — two identical queries on every dequeue. The records that did arrive are now passed in rather than re-queried: one extra query per kind instead of two. That assertion existed for a different reason and caught this for free.

## Controls, each observed

| Break | Result |
|---|---|
| resolve both sides with the enabled-only lookup | 2 tests fail — the reasons become indistinguishable |
| remove the empty-list guard | the "clean run records no step" test fails |

11 new tests: 4 on the step, 7 on the detection against real rows — including that a `hidden` but enabled source is **not** reported (the repositories ignore enable fields, so `hidden` is FormEngine visibility; reporting it would send an operator to fix something that is not broken), and that a deleted record reads `gone` rather than `deactivated`.

## Two corrections to records, both mine

**ADR-179 claimed `RunAugmentation` is `@internal`.** It is on the frozen surface — `api-surface.txt` lists it with its full constructor. Corrected in the record rather than left for the next reader. The growth is additive (a parameter with a default plus one method) and announced under `### Added`.

**ADR-176 said "five builtins write through the DataHandler".** ADR-180 added a sixth within a day of that record being written. The number is replaced by the command that answers it — `grep -l 'use WritesThroughDataHandlerTrait' …`, the trait every writer uses, rather than a search for the class name, which also matches the two traits themselves and overcounts by two. I wrote that wrong anchor first and checked it before trusting it.

Gates: `phpstan` level 10 clean, `unit` 7267 pass, `fuzzy` pass, `functional -d sqlite` on the touched classes pass, `cgl` and `rector -n` stable at PHP 8.2.

Closes #809

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,27 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   reasons that say nothing about the model, so folding it in would measure the
   gate instead of the answer.
 
+- A forced source a queued run asked for and did not get is now recorded on the
+  run (ADR-179).
+
+  Switching a snippet or skill off between enqueue and start drops it — that is
+  ADR-175's rule and stays — but until now nothing said so, and the only way to
+  notice was to compare the queued request against the transcript.
+
+  The run now carries a `dropped` step naming each source and why: `deactivated`
+  for a record that is switched off, `gone` for one that no longer resolves.
+  Those stay apart on purpose — a deactivated record can be switched back on, a
+  removed one cannot, so a single "dropped" would send the reader looking.
+
+  A run that dropped nothing records no step, so the step's presence is the
+  signal. The resume path is untouched: ADR-166 and ADR-175 keep a deactivated
+  source resolving there, so nothing is dropped and a report would imply
+  otherwise.
+
+  `RunAugmentation` gains a `droppedSources` parameter, appended with a default;
+  `RunTrace` gains `recordDroppedSources()`. Both are additive.
+
+
 ### Changed
 - **AGENTS.md files synchronized with the repository state.** Root slimmed from 356 to 119 lines by moving content into the scoped files it belongs to; stale inventories refreshed (TCA files, database tables, backend templates, JS modules, `Services.Dashboard.php`); phantom `TCA/Overrides/` and dead `MEMORY.md` references removed; generic workflow boilerplate in `.github/workflows/AGENTS.md` replaced with this repository's actual conventions (no local jobs, release flow, dependency automation).
 
--- a/Classes/Domain/Enum/DroppedSourceReason.php
+++ b/Classes/Domain/Enum/DroppedSourceReason.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ * Copyright (c) 2025-2026 Netresearch DTT GmbH
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+declare(strict_types=1);
+
+namespace Netresearch\NrLlm\Domain\Enum;
+
+/**
+ * Why a forced source the run asked for did not arrive (ADR-179).
+ *
+ * The two cases are kept apart deliberately. Both resolve to nothing, so a
+ * single "dropped" would be easier to produce — and it would flatten two
+ * different operator actions with different remedies: a deactivated record can
+ * be switched back on, a removed one cannot. A reader who cannot tell them
+ * apart has to go looking, which is the situation this record exists to end.
+ *
+ * @internal Not part of the @api surface; may change without notice (ADR-127).
+ */
+enum DroppedSourceReason: string
+{
+    /**
+     * The record exists and is switched off.
+     *
+     * Detected by resolving the same uid twice: the existence lookup finds it,
+     * the enabled-only lookup does not. That difference is the whole signal —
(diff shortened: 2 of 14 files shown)


CybotTM:

**An existing test found a real defect in my first version.** `ToolLoopServiceAssemblyOrderTest` pins that the enabled-only lookup runs once per dequeue; my detection called it a second time, doubling that query on every queued run. The records that did arrive are now passed in rather than re-queried. That assertion exists for a different reason and caught this for free — which is the argument for pinning query counts at all.

**Both new assertions were seen to fail**, not just seen to pass: resolving both sides with the enabled-only lookup makes the two reasons indistinguishable and fails two tests; removing the empty-list guard makes a clean run record a step and fails the test that forbids it.

**Two of the seven detection tests are the ones I would not have written from the ADR alone.** A `hidden` but enabled source is *not* reported — the repositories ignore enable fields, so `hidden` is FormEngine visibility, and reporting it would send an operator to fix something that is not broken. And a deleted record reads `gone`, not `deactivated`, because both lookups keep the deleted restriction; that is the only reason the two reasons stay separable at all.

**Two corrections to records, both mine, both found by opening the file rather than trusting the sentence.** ADR-179 claimed `RunAugmentation` is `@internal`; `api-surface.txt` lists it with its full constructor. ADR-176 said five builtins write through the DataHandler; ADR-180 added a sixth within a day, so the number is replaced by the command that answers it — and my first anchor for that returned 8, because the two traits mention the class name too. I checked the anchor before trusting it.

**Deliberately not flattened to a count.** `snippet#41 deactivated` says what to do about it; `2 dropped` erases the distinction the record exists to make.

Gates: `phpstan` level 10 clean, `unit` 7267 pass, `fuzzy` pass, `functional -d sqlite` on the touched classes pass, `cgl` and `rector -n` stable at PHP 8.2.

CybotTM:
Self-review: 137c496d74f6e9f71b3b0cebb14b2e25f0558d0b

Issue #848, opened 2026-08-20
MCP import refuses a whole tool for a union in one property — decide whether anyOf/oneOf can be carried


CybotTM:
The MCP import refuses any schema carrying `anyOf`, `oneOf`, `$ref` and their siblings, at any level. The reasoning in `McpSchemaNormalizer` is sound — dropping such a keyword widens what the tool accepts, and a model would then produce arguments the server rejects. But refusing the whole tool is not the only way to honour it, and the cost is visible on the first real server anyone points at this.

**Observed on the demo instance.** `https://mcp.deepwiki.com/mcp` advertises three tools. Two import. `ask_question` — the one an operator actually wants — is refused, because a single property allows either shape:

```json
"repoName": {
  "anyOf": [
    {"type": "string"},
    {"items": {"type": "string"}, "type": "array"}
  ],
  "description": "GitHub repository or list of repositories (max 10) in owner/repo format."
}
```

Everything else about that schema is plain: `question` is a string, both are required, the object is two levels deep. The union sits in one property and describes a convenience, not a subtle constraint.

**Worth deciding, rather than leaving implicit:**

1. **Pass the keyword through unchanged.** The providers this extension talks to accept `anyOf` in a function-call schema (OpenAI outside strict mode, Anthropic passes JSON Schema through). Carrying a keyword verbatim is not the same as dropping it: the risk the current rule guards against is *removal*, not presence. What would need checking is the strict-mode subset per provider and whether the schema still round-trips.
2. **Refuse only where the keyword changes what is accepted.** A union at the top level is a different thing from a union inside one property; `$ref` into a `$defs` block that is not carried is a real dead end, `anyOf` over two inline types is not.
3. **Keep refusing, and say so in the UI as a supported limitation** — with the reason named at the tool (which the sibling change to the import report now does) and, ideally, a note that the server can advertise a flattened schema.

Option 3 is where the code effectively is today; it is defensible, but it should be a decision on the record rather than a side effect of one keyword list.

Note this is not about relaxing the guard for convenience: the current behaviour silently costs a working tool on a widely used public MCP server, and an operator has no way to influence a remote schema.

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_0144iD1P22LotW8rxmxrNGro)_

PR #849, opened 2026-08-20
fix(mcp): say why a tool was skipped on import

CybotTM:
Reported from the demo instance: the MCP module's last import error read

> `"ask_question" has no usable parameter schema and was skipped.`

That sentence is true and gives the operator nothing. It does not separate a malformed schema from an oversized one from a well-formed one this import deliberately refuses, and there is no way to tell which of the three happened without reading the normaliser.

**What it says now:** the reason, and for the common case the keyword whose removal would widen what the tool accepts:

> `"ask_question" was skipped: it uses "anyOf", which this import does not carry: dropping the keyword would widen what the tool accepts, letting a model produce arguments the server then rejects.`

**Verified against the real server.** `https://mcp.deepwiki.com/mcp` declares `ask_question`'s `repoName` as `anyOf[string, array<string>]`; its two sibling tools import fine, which is exactly the "two imported DeepWiki tools" the demo seeds. The new unit test carries that schema verbatim rather than a constructed one.

**The rejection rules are unchanged.** `rejectionReason()` mirrors `normalise()` step for step, in the same order, on the same data — including that the keyword walk runs over the FILTERED schema. A keyword sitting in a key that gets dropped anyway (an `examples` block, say) is therefore never reported as the cause; there is a test for that, because a plausible-but-wrong reason would be worse than the vague one it replaces.

Whether the union should be refused at all is a separate question, filed as #848 rather than decided here.

Tests: three new unit tests (real schema refused with the keyword named, a passing sibling schema reports no reason, a keyword in a dropped key is not blamed). Gates run locally: cgl, PHPStan level 10, the normaliser unit suite, and the functional `McpImport` suite on sqlite — all green; CI covers the eight-cell matrix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Changed
+
+- **An MCP tool that is skipped on import now says why.** The report used to state that a tool had "no usable parameter schema", which is true and unactionable: an operator could not tell a malformed schema from an oversized one from a well-formed one this import deliberately does not carry. It now names the reason — for the common case, the keyword (`anyOf`, `$ref`, …) whose removal would widen what the tool accepts. The rejections themselves are unchanged.
+
 ## [0.31.1] - 2026-08-20
 
 ### Changed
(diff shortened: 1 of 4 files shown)


CybotTM:
Self-review: 6c84fdc9bcb41ffa30df5f474c5a48e4b80d4a5e

PR #852, opened 2026-08-20
fix(telemetry): carry the caller source across every options rebuild

### LINKED ISSUE #845: Feature services drop withCallerSource() before dispatch (ADR-177 unreachable via VisionService, TranslationService, ToolLoopService)
`AbstractOptions::withCallerSource()` (ADR-177) reaches telemetry only on the `LlmServiceManager` entry points. The three feature services that sit in front of it drop the caller identity before dispatch, so an annotated call through `VisionService`, `TranslationService` or `ToolLoopService` is still recorded as unattributed. Found in v0.31.1 while annotating `t3_cowriter` ([t3x-cowriter#161](https://github.com/netresearch/t3x-cowriter/issues/161)).

## Where it is lost

**`VisionService::analyzeImageFull()`** rebuilds the options object field by field before calling `$this->llmManager->vision()`. `callerSourceExtension` / `callerSourceOperation` are protected properties set by a wither, not constructor parameters, so the rebuild cannot carry them and does not:

```php
$visionOptions = new VisionOptions(
    maxTokens: $options->getMaxTokens(),
    temperature: $options->getTemperature(),
    provider: $options->getProvider(),
    model: $options->getModel(),
    beUserUid: $options->getBeUserUid(),
    plannedCost: $options->getPlannedCost(),
);
```

This is the same class of bug the comment right above that block warns about for the budget fields — a missing field looks like "use the default" to PHPStan.

**`TranslationService::translate()`** rebuilds `ChatOptions` the same way. `translateForConfiguration()` and `translateWithTranslator()` route through `chatWithConfiguration()` with `budgetMetadata($options)`, which only ever emits the two `BudgetMiddleware` keys.

**`ToolLoopService`** passes `budgetMetadata($options)` into both `chatWithConfiguration()` calls (per-iteration and cap-hit synthesis); it also only emits the `BudgetMiddleware` keys.

## Reproduction

Against v0.31.1, with a mocked `LlmServiceManagerInterface` capturing the options it receives:

```php
$service = new VisionService($manager);
$service->analyzeImageFull(
    'https://example.com/a.png',
    'prompt',
    VisionOptions::altText()->withCallerSource('t3_cowriter', 'altText'),
);
// captured options: getCallerSourceExtension() === null, getCallerSourceOperation() === null
```

```php
$service->translate(
    'hello',
    'de',
    'en',
    (new TranslationOptions())->withCallerSource('t3_cowriter', 'translate'),
);
// captured ChatOptions: getCallerSourceExtension() === null, getCallerSourceOperation() === null
```

Both print `NULL` / `NULL`.

## Why it matters for ADR-177

ADR-177 chose the options object over the `$metadata` array precisely because "the `$metadata` array on the `*WithConfiguration()` methods is positional plumbing the feature services do not expose". The options path is currently the only channel a feature-service caller has, and it does not arrive — so vision, translation and tool calling cannot be attributed at all, by any means available to a consumer. The `LlmServiceManager` entry points are unaffected.

## Suggested direction (not a spec)

Carrying the two fields across every rebuild is the mechanical fix, but it is the third place the same rebuild has lost a field. A copy helper on `AbstractOptions` for the non-provider call metadata (idempotency key, correlation id, caller source), or building the new options object from a clone of the old one, would close the class rather than this instance. `budgetMetadata()` in `TranslationService` and `ToolLoopService` would gain the `callerSource()` term alongside it.

Whichever shape is chosen, an assertion per feature service that the annotation survives the hop would keep it closed.

### LINKED ISSUE #846: Caller source is dropped on the tool loop's tool-free completions
`ToolLoopService` forwards the caller source only on the tool-calling wire. `chatWithToolsForConfiguration()` receives the `ToolOptions` object itself (line 336), so `LlmServiceManager` adds `CallMetadataFactory::callerSource()` to the middleware context and `TelemetryMiddleware` persists `source_extension` / `source_operation`. The two `chatWithConfiguration()` calls in the same class do not: they pass `$this->budgetMetadata($options)` (lines 284 and 464), and that helper builds its array from `getBeUserUid()` and `getPlannedCost()` only — the caller source never enters it.

Two rounds of a correctly annotated run therefore write an empty origin:

- Line 284 — the loop with no offered tool specs. This is the whole call for a configuration whose tool set resolves empty, and for a resume continuation that offers no tools.
- Line 464 — the synthesis completion after the iteration cap is hit. That is the round that produces the final answer the user reads, on the run's largest transcript, so it is not the cheap one.

The result is a run that shows up in the Analytics breakdown under its extension for its tool rounds and under *Unattributed* for these, i.e. one run split across two rows.

Fix shape: have `budgetMetadata()` (or a wider helper next to it) also emit `TelemetryMiddleware::METADATA_SOURCE_EXTENSION` / `METADATA_SOURCE_OPERATION` from `$options->getCallerSourceExtension()` / `getCallerSourceOperation()`, the same pair `CallMetadataFactory::callerSource()` builds — so both wires agree.

Found while annotating netresearch/t3x-nr-mcp-agent#134 (PR netresearch/t3x-nr-mcp-agent#135); verified by reading nr-llm v0.31.1 as installed.

CybotTM:
The caller identity from ADR-177 only reaches the middleware when the options object the caller annotated is the one that arrives. Three places rebuilt it on the way and dropped the identity, so the per-extension breakdown (ADR-178) reported *Unattributed* no matter how carefully the caller named itself.

**What was losing it:**

- `VisionService::analyzeImageFull()` rebuilds `VisionOptions` from a fixed parameter list to strip `detail_level`. The caller source is set by a wither, not a constructor parameter, so it was invisible to that list — the same trap the comment right above that rebuild already warns about for the budget fields, one field further on.
- `TranslationService` builds `ChatOptions` from `TranslationOptions` at three sites.
- `ToolLoopService` places two completions itself (the tool-free turn and the post-cap synthesis) through `chatWithConfiguration()`, whose channel is the metadata array; `budgetMetadata()` emitted only the two budget keys, so an annotated tool run split across two rows — the tool-capable turns attributed, these two not.

**Why it matters beyond tidiness:** nr-llm-compat's `ai_filemetadata` bridge routes its vision calls through exactly `analyzeImageFull()`, and making third-party AI usage attributable is the whole point of that bridge. It was annotating into a void. The same held for every consumer being annotated right now (nr-landingpage, nr-mcp-agent, nr-repurpose, t3-cowriter, nr-ai-search) on their vision and translation paths.

Found by two independent subagents while annotating consumers, then verified in the source here.

**Tests** — the vision guard was watched failing before it was kept: removing the copy again makes `analyzeImageFullPreservesTheCallerSourceThroughTheRebuild` fail on the options object handed to the manager, and it passes once restored. It sits next to the existing budget-field regression test, which is the same defect class one field earlier.

A second test pins the tool-loop metadata, and writing it caught a real bug in this change: the getter is nullable while the wither writes `''` for an omitted operation, so a first version wrote an empty `sourceExtension` key for unannotated runs. Both cases are asserted now.

Gates run locally: cgl, PHPStan level 10, and the three touched unit suites (`VisionServiceTest`, `TranslationServiceTest`, `ToolLoopServiceTest`). CI covers the eight-cell matrix.

Closes #845, closes #846.

### DIFF
--- a/Classes/Service/Feature/TranslationService.php
+++ b/Classes/Service/Feature/TranslationService.php
@@ -99,14 +99,14 @@ final readonly class TranslationService implements TranslationServiceInterface
             ChatMessage::user($prompt['user']),
         ];
 
-        $chatOptions = new ChatOptions(
+        $chatOptions = $this->carryCallerSource(new ChatOptions(
             temperature: $options->getTemperature() ?? 0.3,
             maxTokens: $options->getMaxTokens() ?? 2000,
             provider: $options->getProvider(),
             model: $options->getModel(),
             beUserUid: $this->resolveBeUserUid($options),
             plannedCost: $options->getPlannedCost(),
-        );
+        ), $options);
 
         $response = $this->llmManager->chat($messages, $chatOptions);
 
@@ -305,14 +305,14 @@ final readonly class TranslationService implements TranslationServiceInterface
                 ['temperature' => 0.1, 'max_tokens' => 10],
             );
         } else {
-            $chatOptions = (new ChatOptions(
+            $chatOptions = $this->carryCallerSource((new ChatOptions(
                 temperature: 0.1,
                 maxTokens: 10,
                 provider: $options->getProvider(),
                 model: $options->getModel(),
                 beUserUid: $this->resolveBeUserUid($options),
                 plannedCost: $options->getPlannedCost(),
-            ))->withSuppressRequestCount(!$countsAsRequest);
+            ))->withSuppressRequestCount(!$countsAsRequest), $options);
 
             $response = $this->llmManager->chat($messages, $chatOptions);
         }
@@ -356,14 +356,14 @@ final readonly class TranslationService implements TranslationServiceInterface
             )),
         ];
 
-        $chatOptions = new ChatOptions(
+        $chatOptions = $this->carryCallerSource(new ChatOptions(
             temperature: 0.1,
             maxTokens: 10,
             provider: $options->getProvider(),
             model: $options->getModel(),
             beUserUid: $this->resolveBeUserUid($options),
             plannedCost: $options->getPlannedCost(),
-        );
+        ), $options);
 
         $response = $this->llmManager->chat($messages, $chatOptions);
 
@@ -675,5 +675,22 @@ final readonly class TranslationService implements TranslationServiceInterface
     private function resolveBeUserUid(TranslationOptions $options): ?int
     {
         return $options->getBeUserUid() ?? $this->beUserContextResolver?->resolveBeUserUid();
+    }
+
+    /**
+     * Carry the caller identity across an options rebuild.
+     *
+     * Every `new ChatOptions(...)` in this service builds a fresh object from a
+     * fixed parameter list, and the caller source is set by a wither rather than
+     * a constructor parameter — so without this it is silently dropped and the
+     * call lands in the "Unattributed" bucket of the per-extension breakdown
+     * (ADR-177/ADR-178), no matter what the caller annotated.
+     */
+    private function carryCallerSource(ChatOptions $rebuilt, TranslationOptions $source): ChatOptions
+    {
+        return $rebuilt->withCallerSource(
+            $source->getCallerSourceExtension() ?? '',
+            $source->getCallerSourceOperation() ?? '',
+        );
     }
 }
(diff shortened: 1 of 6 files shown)


CybotTM:
Self-review: d3ab404f0ad770559863240b472dc3b450013692

PR #853, opened 2026-08-20
feat(mcp): carry union keywords to the provider instead of skipping the tool

### LINKED ISSUE #848: MCP import refuses a whole tool for a union in one property — decide whether anyOf/oneOf can be carried

CybotTM:
Implements the decision on #848: carry the union, do not skip the tool.

**The rule that was there is sound, but it answered a different question.** Refusing `anyOf` protects against *dropping* it — a stored schema without the union permits more than the server does, so a model can produce arguments the server then rejects. Carrying the keyword unchanged does not do that: `ToolSpec::toArray()` puts the stored schema into the provider payload verbatim, and the providers this extension talks to accept a union in a function schema. The narrowing the server declared survives end to end.

`anyOf`, `oneOf`, `allOf` and `not` are therefore carried now, and added to the retained top-level keys so a union declared at the root survives the filter too (inside a property it rides along with the property schema, which is copied whole).

**What stays refused, and why the line runs exactly there:**

- `$ref`, `$dynamicRef`, `$defs`, `definitions` — a reference points into a definition block this filter does not carry, so carrying the pointer alone would hand the provider a dangling reference. Refusing is the honest answer until the definition block is carried too.
- `if`/`then`/`else`, `dependentSchemas`, `dependentRequired`, `unevaluatedProperties`, `patternProperties`, `propertyNames` — the draft-2019/2020 applicators have no dependable support across providers. Carrying them would trade an import-time refusal, which names the keyword since #849, for a call-time failure with a worse message.

**The case that prompted this.** `https://mcp.deepwiki.com/mcp` advertises three tools; `ask_question` declares `repoName` as `anyOf[string, array<string>]` and was the one skipped, while its two siblings imported. That is now the test: the schema is carried verbatim from the real server's response, and the assertion checks the `anyOf` block arrives unchanged rather than merely that the tool imports.

**Tests:** the real-world union imports and keeps its `anyOf`; a top-level union survives the key filter; a `$ref` schema is still refused and its reason names the keyword. Two existing data-provider cases asserted the old behaviour for `allOf` and nested `oneOf` — they are replaced by cases for the applicators that remain refused, so the provider still pins a real boundary rather than shrinking.

Gates run locally: cgl, PHPStan level 10, the normaliser unit suite, and the functional `McpImport` suite on sqlite. CI covers the eight-cell matrix.

Closes #848.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- **An MCP tool whose schema uses a union now imports.** `anyOf`, `oneOf`, `allOf` and `not` are carried to the provider verbatim instead of causing the whole tool to be skipped. The rule they were caught by protects against *dropping* a constraint — which would let a model produce arguments the server rejects — and carrying one unchanged does not do that. References (`$ref`, `$defs`) and the draft-2019/2020 applicators (`if`/`then`/`else`, `dependentRequired`, `unevaluatedProperties`, …) stay refused: the first would arrive dangling because the definition block is not carried, the second have no dependable provider support. Concretely, DeepWiki's `ask_question` imports now (#848).
 - **An MCP tool that is skipped on import now says why.** The report used to state that a tool had "no usable parameter schema", which is true and unactionable: an operator could not tell a malformed schema from an oversized one from a well-formed one this import deliberately does not carry. It now names the reason — for the common case, the keyword (`anyOf`, `$ref`, …) whose removal would widen what the tool accepts. The rejections themselves are unchanged.
 
 ## [0.31.1] - 2026-08-20
--- a/Classes/Service/Tool/Mcp/McpSchemaNormalizer.php
+++ b/Classes/Service/Tool/Mcp/McpSchemaNormalizer.php
@@ -49,6 +49,10 @@ final readonly class McpSchemaNormalizer
      * The only top-level keys carried over. Everything else is either an
      * annotation a provider ignores (`title`, `$id`, `examples`, vendor
      * extensions) or a constraint handled by self::UNSUPPORTED_KEYWORDS.
+     *
+     * The applicators are listed so a union declared at the TOP level survives
+     * the filter; inside a property they ride along with the property schema,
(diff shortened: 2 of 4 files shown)


CybotTM:
Self-review: fd1a0bd4bea855ea0c3993a08f62fb90ea701135

Issue #854, opened 2026-08-20
Twelve innerHTML assignments the sanitizer rule cannot see through escapeHtml()


CybotTM:
Switching ESLint on (#825) reports 12 `no-unsanitized/property` findings across five backend scripts. **None of them is a hole**, and that is the finding: the code escapes, the rule cannot see it, and the two are now permanently out of step.

## The evidence

`SetupWizard.js:471-474` is the clearest case — the escaping is deliberate, commented, and sits immediately above the assignment the rule flags:

```js
// SECURITY: Escape all external data to prevent XSS
const safeProviderName = escapeHtml(providerInfo.suggestedName);
const safeAdapterType  = escapeHtml(providerInfo.adapterType);
const safeEndpoint     = escapeHtml(providerInfo.endpoint);
```

`escapeHtml()` comes from `@netresearch/nr-llm/Backend/HtmlEscape.js`. `no-unsanitized` recognises a fixed set of sanitizers and a helper of ours is not among them, so every assignment downstream of it reads as unsafe.

## Where they are

| File | Lines |
|---|---|
| `SetupWizard.js` | 268, 295, 385, 404, 476, 496, 506 |
| `TaskExecute.js` | 215, 247 |
| `ConfigurationList.js` | 251 |
| `ModelList.js` | 102 |
| `ProviderList.js` | 76 |

Two of them — `TaskExecute.js:215` and `:247` — are a different shape again: a saved `innerHTML` restored onto the same element after a spinner. That round-trips a value the DOM already held; it injects nothing new.

## Why this is its own issue and not a config comment

#825 exists to catch the defect class that passed 71 green checks — a method that landed outside its class. `no-undef` catches that, and it is an **error** in the config that ships with #825.

The sanitizer rule is set to `warn` there instead, for a measured reason rather than a cautious one: as an error it lands twelve judgements on a change whose job is something else. Dropping the plugin entirely is not an option either — two files already carry `eslint-disable-line no-unsanitized/property`, suppressions written for a rule that never ran, which become "rule not found" errors the moment linting is switched on.

**A security question does not belong in a config comment**, which is why it is here where it can be argued and closed.

## Options

1. **Teach the rule about `escapeHtml()`** via the plugin's `escape` option, then raise both rules to `error`. Cleanest, and the suppressions in the two files can go.
[...]

PR #855, opened 2026-08-20
ci(js): lint the backend JavaScript, which nothing did

### LINKED ISSUE #825: Nothing checks the backend JavaScript, and a broken file passed the whole matrix
The extension ships ~20 backend JavaScript modules under `Resources/Public/JavaScript/`. **No gate reads any of them.** There is no eslint config, no `node --check`, and no lint step in CI or `Build/`. The Playwright suite exists but covers other views.

## How it surfaced

In #823 a set of methods was appended to `TaskExecute.js` after the class's closing brace — they landed inside the DOM-ready `if/else` instead of the class body. `this.showOutcomeBar(...)` was therefore a call to a method that does not exist on the class, so the rating buttons the PR is about would not have worked.

**The full CI matrix was green: 71 checks passing.** PHPStan does not read JavaScript, the functional suite renders Fluid but never executes the module, and no Playwright spec touches the AI-tasks execute view. The defect was found by a person reading the diff.

## What a check would have to catch

Not only a parse error. A syntax check alone is not obviously enough here: `node --check` accepted the broken file while a minimal reproduction of the same shape was rejected, and I could not reconcile the two. What is certain and check-independent is structural — the class closed at one line and the methods began after it, so they were never class members.

That points at a rule about scope and undefined references rather than a bare parser run:

- `no-undef` / `no-unused-private-class-members` style rules catch a method that is called but not defined on the class.
- A parser run is cheap and worth having anyway, but on its own it may not have failed here.

## Options

1. **eslint with a small rule set**, wired into `ci:test:repo` so it runs in the shared `repo-checks` job and in pre-commit. Most coverage, needs a config and a dependency this repo does not have today.
2. **A Playwright spec per interactive view.** Catches this and more, but the suite is slow and the failure mode is a timeout rather than a named cause.
3. **`node --check` over the directory.** Cheapest, no dependency — and per the above, not certain to have caught this one. Worth it as a floor, not as the answer.

Option 1 addresses the class of defect; option 3 is a floor that could ship in an hour. They are not exclusive.

## Scope note

Whatever is chosen must run in CI, not only in pre-commit: the file in #823 was committed from a worktree with no hooks installed, which is the normal case for a fresh worktree in this repository.

Found while reviewing #823. Ref ADR-176.

CybotTM:
Closes #825. A file whose methods had landed outside their class passed the whole CI matrix — **71 green checks** — and the rating buttons it shipped would not have worked. PHPStan does not read JavaScript, the functional suite renders the template without executing the module, and no Playwright spec covers that view. A person reading the diff found it.

## `node --check` lost on evidence, not on taste

It was the cheap candidate: no dependency, no config. Tested against the actual broken file, it **accepted** it — because it parses a `.js` file as a CommonJS *script*. ESLint parses it as an ES *module*, which is how the browser loads it, and reports the parse error on the right line.

That also settles something left open in #842's review. I had written there that I could not reconcile `node --check` passing with a minimal reproduction failing, and would not dress it up. The answer is the parsing mode: the file was a syntax error all along, and `node --check` read it in the wrong one.

## My reason for not deciding this earlier was wrong

I had said eslint "would bring a JS toolchain into a PHP extension that has none". One command disproves it: `package.json`, `package-lock.json`, three devDependencies including Playwright, and CI already runs `npm install` for the e2e job. ESLint is one more line in a file that exists.

**No sibling extension lints JavaScript** — `rte_ckeditor_image` has a `package.json` for commitlint and husky only, `nr-vault` one for Playwright, `nr-repurpose` none at all. There was no house standard to copy, which is a reason to set one rather than to wait for it.

## The rule set is small on purpose

`no-undef` is an **error**: it is the rule that catches the defect above.

`no-unsanitized` is loaded but set to **warn**, and the reason is measured rather than cautious. As an error it reports 12 assignments, and **none of them is a hole** — the code escapes through `escapeHtml()`, visibly, with a comment saying so:

```js
// SECURITY: Escape all external data to prevent XSS
const safeProviderName = escapeHtml(providerInfo.suggestedName);
```

The rule recognises a fixed set of sanitizers and a helper of ours is not among them. Dropping the plugin was not an option either: two files already carry `eslint-disable-line no-unsanitized/property` — suppressions written for a rule that never ran, which become "rule not found" errors the moment linting is switched on.

Those twelve are triaged in **#854** with the evidence and three options. A security question belongs where it can be argued and closed, not in a config comment.

## Verification

| | |
|---|---|
| clean tree | 0 errors, 14 warnings, exit 0 |
| #825's shape reintroduced (method after the class's closing brace) | parse error, exit 1 |
| `npm ci` from the lockfile, then lint | exit 0 |

Repo rules: no job is defined here — thin caller of the shared `script-check` reusable, `check-workflow-ownership.php` passes, and `drift_compare.py` against the org template reports no drift. `pre-command` installs the devDependencies the runner does not ship.

## One thing worth knowing beyond this repo

This makes nr_llm the first Netresearch TYPO3 extension with a JavaScript gate. If it holds up, it belongs in the shared template rather than being rebuilt per repository — otherwise the next extension gets the same 71-green-checks surprise.

### DIFF
--- a/.github/workflows/js-lint.yml
+++ b/.github/workflows/js-lint.yml
@@ -1,0 +1,34 @@
+name: JavaScript Lint
+
+# Lints Resources/Public/JavaScript. Nothing did before #825: a file whose
+# methods had landed outside their class passed the whole CI matrix — 71 green
+# checks — because PHPStan does not read JavaScript, the functional suite
+# renders the template without executing the module, and no Playwright spec
+# covers that view. The defect was found by a person reading the diff.
+#
+# `node --check` was the cheaper candidate and lost on evidence: it parses a
+# `.js` file as a CommonJS script and accepted the broken file, while ESLint
+# parses it as an ES module — which is how the browser loads it — and reports
+# the parse error on the right line.
+#
+# Thin caller of the shared script-check reusable, so the checkout pin and
+# harden-runner stay maintained centrally. `pre-command` installs the npm
+# devDependencies the runner does not ship.
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  merge_group:
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  js-lint:
+    uses: netresearch/.github/.github/workflows/script-check.yml@main
+    permissions:
+      contents: read
+    with:
+      pre-command: npm ci --no-audit --no-fund
+      check-cmd: npx eslint
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,0 +1,62 @@
+// Lints the backend JavaScript. Nothing did before: a file whose methods had
+// landed outside their class passed the whole CI matrix — 71 green checks —
+// because PHPStan does not read JavaScript, the functional suite renders the
+// template without executing the module, and no Playwright spec covers that
+// view (#825).
+//
+// `node --check` was considered as a cheaper floor and rejected on evidence:
+// it parses a `.js` file as a CommonJS script and accepted the broken file,
+// while ESLint parses it as an ES module — which is how the browser loads it —
+// and reports the parse error on the right line.
+//
+// The rule set is deliberately small. This is the first JavaScript gate in any
+// of our TYPO3 extensions, and a large one would arrive as a backlog of
+// pre-existing violations that nobody asked for.
+import globals from 'globals';
+import noUnsanitized from 'eslint-plugin-no-unsanitized';
+
+export default [
+    {
+        files: ['Resources/Public/JavaScript/**/*.js'],
+        // Vendored libraries are shipped as-is; linting them would report
+        // somebody else's code in our gate.
+        ignores: ['Resources/Public/JavaScript/Vendor/**'],
+        languageOptions: {
+            ecmaVersion: 2022,
+            sourceType: 'module',
+            globals: {
+                ...globals.browser,
+                // Provided by the TYPO3 backend at runtime, not importable.
+                TYPO3: 'readonly',
+                bootstrap: 'readonly',
+            },
+        },
+        plugins: { 'no-unsanitized': noUnsanitized },
+        rules: {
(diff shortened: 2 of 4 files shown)


CybotTM:

**The check I would not have thought to make, and did:** `js-lint / check: pass` appears in this PR's own check list. A new gate that is configured but never triggers is the declaration-nothing-reads failure in its purest form, and "the PR is green" would have hidden it — the other 74 checks are green with or without it.

**The rule choice is measured, not preferred.** `node --check` was the cheaper candidate and I tested it against the actual broken file from #842: it **accepted** it, because a `.js` file parses as a CommonJS script there. ESLint parses it as an ES module — the mode the browser uses — and reports the parse error on the right line. That also closes a question I explicitly left open in #842's review rather than guessing at.

**The control was run in both directions.** Clean tree: 0 errors, exit 0. #825's shape reintroduced — a method after the class's closing brace — parse error, exit 1. And `npm ci` from the committed lockfile followed by a lint run: exit 0, so the gate does not depend on my local `node_modules`.

**Twelve sanitizer findings are deliberately warnings, and the reason is in #854, not here.** None of them is a hole: the code escapes through `escapeHtml()` with a comment saying so, and the rule cannot recognise a helper of ours as a sanitizer. Loading the plugin anyway is not optional — two files carry `eslint-disable-line no-unsanitized/property` for a rule that never ran, and those become errors the moment linting starts.

**One correction to my own earlier reasoning, stated in the PR body too:** I had declined to decide this on the grounds that eslint would introduce a JS toolchain into a PHP extension. `package.json`, a lockfile and three devDependencies were already there. One command would have told me.

Repo rules: no job defined here (thin caller of the shared reusable), `check-workflow-ownership.php` passes, template drift reports none.

Merging on that basis.

CybotTM:
Self-review: 633798d7475c691e4b330c03079c5a5847569085

PR #858, opened 2026-08-20
docs(adr): cite symbols, not line numbers, and check what a test can check

CybotTM:
Refs #793 — option 2 from the issue, applied to the record that carried the most citations, plus the convention that stops the next one reinventing them.

## The citation the existing guard passes

#807 built the cheap half deliberately and said so in its own docblock: it catches a citation naming a file that is gone, or pointing past the end of a file, or at a blank line — and *not* a line that moved onto different but non-blank code, "which is the argument for citing a symbol rather than a line number wherever the prose allows it."

That case is in the corpus:

| | |
|---|---|
| ADR-171 said | `AgentRunController.php:267` — "the list viewport" |
| Line 267 is | `    }` — the closing brace of `submitInputAction()` |
| The list viewport is | `listAction()`, at :119 |
| `AdrCodeCitationTest` said | pass, on all four assertions |

## What moved the citations, exactly

[`9b902993`](https://github.com/netresearch/t3x-nr-llm/commit/9b902993) — "docs(adr): correct the module inventory ADR-119 opens with" — added **+10 lines** near the top of ADR-119. Three citations into that record, from two different ADRs, moved off their targets in one commit and nothing failed:

- ADR-171 → `Adr119:69`, quote is at :79
- ADR-171 → `Adr119:79`, quote is at :89
- ADR-169 → `Adr119:94`, the Decision is at :104

A record correcting one of its own counts silently invalidated three other records' references to it. That is the mechanism, not an accident of one commit.

## Three forms, chosen by what the citation can carry

Not every citation converts the same way, and a surprising share of them converted to *nothing*:

1. **Delete.** The symbol was already named in the same sentence. `:php:`hasGrant()`, :php:`mayAccessSession()` and :php:`mayActOnRun()` (``AiActorContext.php:180``, ``:196``, ``:217``)` — the line numbers add no information and rot anyway.
2. **A symbol.** `ResumeCoordinator.php:205` → `:php:`ResumeCoordinator::approve()``. Survives every edit that does not rename it.
3. **A `:ref:` into the quoted record**, with the quote left in the prose. `AdrLifecycleTest` already fails on a target that does not resolve, so these get a guard for free.
4. **`path#anchor`** — the file plus a unique string it must still contain — for configuration arrays, Fluid templates and XLIFF, which have no symbols. `Configuration/Backend/Modules.php#nrllm_aitasks`, `ApprovalControls.html#argumentsJson`.

Form 4 is **the only citation shape in this corpus a test can hold to its claim**, which is why it exists rather than just dropping to a bare filename.

## The parser was reading 72% of the corpus

The regex required `.php`. ADRs cite templates, XLIFF, the frozen `api-surface.txt` and *each other* by line just as readily:

| form | count | was checked |
|---|---|---|
| `.php` | 67 | yes |
| bare class name (`AiTaskController:106`) | 11 | no |
| `.rst` | 10 | no |
| `.html` | 3 | no |
| `.txt`, `.xlf` | 2 | no |

Broadening it cost nothing — the unresolvable `.php` citations are already on the declared vendor list — and immediately turned up two more defects: the blank-line citation in ADR-119 above, and `Adr169:111` citing `Backend/SetupWizard/Index.html`, a **truncated path resolving to no file at all**. The eleven bare class names all lived in ADR-171 and are gone with it.

## Both new assertions were watched failing

A green regression test on a fixed tree proves it runs, not that it catches anything.

| control | result |
|---|---|
| anchor mangled to `wizard.field.pidX` | `Adr169…rst:112 anchors on "wizard.field.pidX", which Resources/Private/Templates/Backend/SetupWizard/Index.html no longer contains` |
| a `.rst:NNN` citation reintroduced onto a blank line | `Adr171…rst:41 cites …Adr119BackendModulePlacement.rst:94, which is blank` — previously not read at all |
| tree restored | `OK (4 tests, 144 assertions)` |

## Two claims in ADR-171 were wrong, found by converting them

Converting a citation means reading what it points at, which is the point:

- **"nine enforcement points in four classes"** is now ten. `TaskExecutionController` gained a third at `:347`. The record now prints the grep next to the number, per the ADR-169 precedent the issue cites.
- **ADR-169's quote of ADR-119** — "keep the modules under Administration for now" — was cited at `:94`, a blank line. It is at :104.

`Adr/Index.rst` gains the convention, including that a countable claim carries the command that counts it. Without it the next record reintroduces exactly what this one removed.

## Verification

| gate | result |
|---|---|
| `-s unit` (full) | 7286 tests, 24719 assertions, exit 0 |
| `-s phpstan` | No errors |
| `-s cgl -n` | SUCCESS |
| `-s rector -n` | 65 files with **and without** this change; `AdrCodeCitationTest` was among them before it was touched. The 8.2-pinned verdict is CI's — this build resolves 8.4. |

## Scope

ADR-171 (41 citations) and the four in ADR-169 that the broadened parser caught. **ADR-169's remaining 43 are not converted here** — it is the corpus's densest citer and deserves its own diff rather than being buried in this one. Its line citations still pass the checks that existed before this PR, and now the broadened ones too.

### DIFF
--- a/Documentation/Adr/Adr169RecordManagementUsesTypo3Permissions.rst
+++ b/Documentation/Adr/Adr169RecordManagementUsesTypo3Permissions.rst
@@ -108,7 +108,8 @@ mount (``:7677``) and the ``perms_*`` bitmask (``:7696-7705``).
      - Permitted today. No TCA change.
 
 **This is not theoretical for this extension.** The setup wizard renders a
-"Storage Folder (Page ID)" input (``Backend/SetupWizard/Index.html:328-331``),
+"Storage Folder (Page ID)" input
+(``Resources/Private/Templates/Backend/SetupWizard/Index.html#wizard.field.pid``),
 whose value reaches ``saveAction`` (``SetupWizardController.php:301``) and is
 written onto the provider (``:372-374``), every model (``:449``) and every
 configuration (``:525``) with no page check. All eight Extbase repositories set
@@ -292,7 +293,7 @@ reservation in ADR-130 and ADR-131 is retired rather than fulfilled.
 -------------------
 
 :ref:`ADR-119 <adr-119>` decided "keep the modules under Administration for
-now" (``Adr119BackendModulePlacement.rst:94``) and already carries the status
+now" (:ref:`ADR-119 <adr-119-decision>`) and already carries the status
 ``Accepted (deferred — the placement is not finally settled, see Revisit)``. Its
 trigger is "the first cross-consumer editor surface … a personal
 usage-and-budget view, a personal run history, or a lead-editor view over a
--- a/Documentation/Adr/Adr171PersonasTheCodeAlreadyAssumes.rst
+++ b/Documentation/Adr/Adr171PersonasTheCodeAlreadyAssumes.rst
@@ -47,13 +47,14 @@ Personas
 ------------------------------------------
 
 **Code.** Fifteen of the sixteen registered backend modules are
-``'access' => 'admin'`` (``Configuration/Backend/Modules.php:53`` … ``:321``);
-forty ``denyNonAdmin()`` call sites in fourteen controllers restore that line on
-AJAX routes (:ref:`ADR-037 <adr-037>`, enumerated in
-``Adr169RecordManagementUsesTypo3Permissions.rst:308-341``). ``isAdmin``
-short-circuits :php:`hasGrant()`, :php:`mayAccessSession()` and
-:php:`mayActOnRun()` (``AiActorContext.php:180``, ``:196``, ``:217``) and bypasses
-the configuration group restriction (``ConfigurationResolver.php:218``). The
+``'access' => 'admin'`` — ``grep -c "'access' => 'admin'"
+Configuration/Backend/Modules.php`` returns 15. Forty ``denyNonAdmin()`` call
+sites in fourteen controllers restore that line on AJAX routes
+(:ref:`ADR-037 <adr-037>`, enumerated with its own grep in
+:ref:`ADR-169 <adr-169-q6>`). ``isAdmin`` short-circuits :php:`hasGrant()`,
+:php:`mayAccessSession()` and :php:`mayActOnRun()`, and bypasses the
+configuration group restriction (:php:`ConfigurationResolver::actorMayUse()`).
+The
 docblock records where that comes from: "Administrators hold every grant
 implicitly (the core ``check()`` short-circuits on isAdmin)"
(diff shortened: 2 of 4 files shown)


CybotTM:

**The check that mattered, and it is not a gate.** A docs PR that renames citations can silently point them somewhere else, and no test in this repo would notice for the two forms that are not `path#anchor`. So every one of the 41 converted citations in ADR-171 was resolved against the tree before it was written: 19 `:php:` symbols verified to exist as `function <name>`, 4 anchors verified with `grep -F`, every `:ref:` target verified against the corpus's actual labels. That pass caught three things I had written and not measured — a grep I claimed returned nine when it returns ten, and two method names (`indexAction`, `executeAction`) that do not exist on `AiTaskController` at all. Had I trusted the diff's plausibility, the record would assert two invented symbols.

**The `AgentRunController.php:267` finding is the load-bearing one and it is verified, not inferred.** :267 is `    }`, the closing brace of `submitInputAction()` which starts at :213. `listAction()` — what the prose calls "the list viewport" — is at :119. All four assertions in `AdrCodeCitationTest` passed on that citation before this change, which is exactly the blind spot #807's own docblock named.

**The +10 drift has a named cause, and my first explanation was wrong.** I wrote "because ADR-119 grew a section header" and then checked: `git show --numstat 9b902993` is +10 on that file, and it is "correct the module inventory ADR-119 opens with" — a corrected count near the top, not a header. The claim in `Adr/Index.rst` was rewritten to the measured cause before commit.

**Both new assertions were watched failing**, in the position they are meant to catch: a mangled anchor reports the file and the missing string; a `.rst:NNN` citation onto a blank line is now reported where the pre-change parser did not read it at all. The tree was restored and re-run green afterwards, and `git status` shows only the four intended files.

**The broadening is not free coverage dressed up as a win — it was measured first.** Simulating it before touching the test showed the 23 unresolvable `.php` citations are already on the declared vendor list, so the only new failures would be genuine. Two were: the ADR-119 blank line and `Adr169:111`'s truncated path.

**Rector**: 65 files with and without this change, and `AdrCodeCitationTest` is among them on the unmodified tree — the `ChangeOrIfContinueToMultiContinueRector` proposal targets a line whose shape predates this PR. This build resolves 8.4; the 8.2-pinned verdict is CI's, and CI is green.

**Scope left undone deliberately**: ADR-169's remaining 43 citations. Stated in the PR body rather than left for a reader to discover.

Not merging on this record — I hold no merge authorisation for this PR.

CybotTM:
Self-review: d37804e13a94839ed3371843eef6ba7a9474e97f

Issue #844, opened 2026-08-20
Specialized services (image, speech, transcription, DeepL) cannot carry caller source — withCallerSource() compiles there and is dropped


CybotTM:
The specialized services — image generation, speech synthesis, transcription, DeepL translation — spend money that the per-extension breakdown (ADR-178) cannot attribute. Their options objects *look* annotatable and are not, which is the worst of both: a consumer can call `withCallerSource()` on them, the call compiles, and the value is silently dropped.

**Two independent reasons, both verified on `main`:**

1. **`fromArray()` drops it.** All four option classes under `Classes/Specialized/Option/` (`ImageGenerationOptions`, `SpeechSynthesisOptions`, `TranscriptionOptions`, `DeepLOptions`) extend `AbstractOptions`, so they inherit `withCallerSource()` and its getters. But the specialized services reach the generators through a plain array and rebuild the object with `fromArray()`, which reads only its own keys (model, size, quality, style, format, configuration, beUserUid, plannedCost for the image case). Anything the base class holds — including the caller source — does not survive the round trip.

2. **Nothing reads it on that path anyway.** `CallMetadataFactory::callerSource()` is the only reader of `getCallerSourceExtension()`, and its only caller is `LlmServiceManager`. The specialized image and speech services call `runLifecycle()` directly rather than going through the manager's middleware pipeline, so even an options object that kept the value would never reach `TelemetryMiddleware` or `UsageMiddleware`.

**Why it matters now.** ADR-178 made per-extension cost readable, and the consuming extensions are being annotated right now (nr-landingpage, nr-mcp-agent, nr-repurpose, t3-cowriter, nr-ai-search). Image generation is among the most expensive single operations an installation can trigger — it is exactly the line item an operator wants attributed, and it is the one that will stay in *Unattributed* while every text call moves out of it. That makes the breakdown quietly misleading rather than merely incomplete.

Found while annotating nr-landingpage: its AI image generation was the one spending path that could not be annotated, so it was deliberately left alone rather than given a declaration nothing reads (netresearch/nr-landingpage#42).

**What a fix needs to decide:** whether `fromArray()`/`toArray()` should carry the base-class identity for all four classes, and whether the specialized services should route their telemetry and usage through the same metadata the middleware reads. The second half is the substantive one — the first alone would still leave the value unread.

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_0144iD1P22LotW8rxmxrNGro)_

[...]

PR #859, opened 2026-08-21
fix(provider): let vision() and embed() use the default configuration

### LINKED ISSUE #851: vision() has no default-provider fallback while chat() does — an image upload on the demo 500s because of it
`LlmServiceManager::vision()` requires the caller to name a provider. `chat()` does not — it resolves the default configuration when the options carry no provider key. So a vision call written the way the documentation invites (options without a provider, "model selection is nr-llm's job") throws instead of running:

```
ProviderException #4867297358: No provider specified and no default provider configured.
… KeyedProviderRegistry.php line 69
```

**Where the two paths diverge** (`Classes/Service/LlmServiceManager.php`, current `main`):

| | line | what it does with a null provider key |
|---|---|---|
| `chat()` | ~560 | `configurationResolver->resolveDefaultConfiguration($providerKey)` — falls back to the installation's default configuration |
| `vision()` | 717 | `$this->getProvider($providerKey)` — hands `null` straight to the registry, which throws |

**Observed in production, not constructed.** On typo3-demo, uploading an image in the AI-chat module answers `HTTP 500`. The instance log has the exception above with `request_url = /typo3/ajax/ai-chat/file-upload`. The chain is: the upload writes a FAL file → `mfd/ai-filemetadata` generates an alt text on upload (the demo enables that) → [nr-llm-compat](https://github.com/netresearch/t3x-nr-llm-compat) intercepts it and routes it through `VisionServiceInterface::analyzeImageFull()` → `VisionService` hands the options on with `provider: null` → `vision()` throws → nothing catches it, so the upload request dies.

The compat bridge builds `new VisionOptions(temperature: …)` and deliberately names no provider: its whole design is that model and provider selection belong to nr-llm rather than to the third-party extension it stands in for. That is the intended usage, and it cannot work today.

The same instance shows the second half of the symptom: with vision resolution failing, `getProviderCapabilities()` in nr-mcp-agent catches the throw and reports `visionSupported: false`, so a *selected* image is never expanded into the message and the assistant answers that it sees no image. One missing fallback, two user-visible failures.

**Suggested fix:** give `vision()` the same resolution `chat()` has — resolve the default configuration when no provider key is supplied — and check the sibling entry points (`image()`, `speech()`, `embed()`) for the same asymmetry rather than fixing only the one that was reported.

**Worth deciding alongside:** whether a provider-resolution failure inside a FAL upload listener should be able to fail the upload at all. A missing alt text is a degraded result; a failed file upload is a broken one.

CybotTM:
Fixes #851: `vision()` and `embed()` now resolve the backend-managed default configuration when the caller pins no provider, which `chat()` has done since ADR-034.

**What was happening.** Both handed the provider key straight to the registry, so a `null` key threw `No provider specified and no default provider configured` — on installations that have a perfectly good default. That is the usage the feature services invite: options without a provider, because model selection is nr-llm's job rather than the caller's.

**Two user-visible failures on typo3-demo came from this one gap.** An image upload in the AI-chat module answered `HTTP 500`, because `mfd/ai-filemetadata` generates an alt text inside the upload request and nr-llm-compat routes that through `VisionServiceInterface`. And with vision resolution failing, `getProviderCapabilities()` reported `visionSupported: false`, so a *selected* image was never expanded into the message and the assistant answered that it saw no image. The instance log carries the exception with `request_url = /typo3/ajax/ai-chat/file-upload`.

**What the fix does, and what it deliberately does not.** The resolved configuration drives the call rather than merely unblocking it: its model reaches the provider (which would otherwise fall back to its own hardcoded default), and the pipeline context carries the real configuration instead of a synthesised ad-hoc one, so budget and telemetry attribute the call to it. The caller keeps precedence — an explicitly pinned provider skips the resolution entirely, and a model named in the options is left alone. With no default configuration, or one that is model-less or access-restricted, the call still refuses: the resolver's existing guards apply unchanged, so this resolves a default rather than inventing one.

`embed()` had the identical asymmetry and is fixed in the same commit; the issue asked for the siblings to be checked rather than only the reported one. The specialized services (image, speech, DeepL) resolve differently and are not touched here — their attribution gap is #844.

**Tests.** Three new cases: the default configuration is used and its model arrives at the provider; a caller-named model survives; and with no default configuration the call still throws. Watched failing before being kept — removing the fallback again reproduces the demo's exception verbatim in two of them.

Gates run locally: cgl, PHPStan level 10, the full unit suite (7288 tests), and the changelog check. CI covers the eight-cell matrix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- **`vision()` and `embed()` use the default configuration when no provider is pinned**, as `chat()` has since ADR-034. Handing them no provider used to throw "No provider specified and no default provider configured" on an installation that has a perfectly good default — which is what the feature services invite, since model selection is nr-llm's job. The resolved configuration drives the call (its model reaches the provider) and shows up in telemetry as itself rather than as an ad-hoc entry; a provider or model the caller named wins, and with no default configuration the call still refuses rather than picking one (#851).
+
 ### Changed
 
 - **An MCP tool whose schema uses a union now imports.** `anyOf`, `oneOf`, `allOf` and `not` are carried to the provider verbatim instead of causing the whole tool to be skipped. The rule they were caught by protects against *dropping* a constraint — which would let a model produce arguments the server rejects — and carrying one unchanged does not do that. References (`$ref`, `$defs`) and the draft-2019/2020 applicators (`if`/`then`/`else`, `dependentRequired`, `unevaluatedProperties`, …) stay refused: the first would arrive dangling because the definition block is not carried, the second have no dependable provider support. Concretely, DeepWiki's `ask_question` imports now (#848).
--- a/Classes/Service/LlmServiceManager.php
+++ b/Classes/Service/LlmServiceManager.php
@@ -619,6 +619,7 @@ final readonly class LlmServiceManager implements LlmServiceManagerInterface, Si
     {
         $options ??= new EmbeddingOptions();
         [$providerKey, $optionsArray] = $this->splitProviderKey($options->toArray());
+        [$configuration, $providerKey, $optionsArray] = $this->applyDefaultConfiguration($providerKey, $optionsArray);
 
         // Cache metadata: EmbedCacheKeyBuilder returns an empty array when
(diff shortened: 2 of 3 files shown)


CybotTM:
Self-review: b4c7a11543b73bdeb4ad692885b05c94cc758f4d

PR #860, opened 2026-08-21
docs(adr): retire every line citation in the corpus, and three of them were wrong

### LINKED ISSUE #793: ADRs cite code by line number and the citations rot silently — nothing checks them

CybotTM:
Closes #793. Stacked on #858 — review that one first; this PR's diff against it is ADR-169, ADR-140, ADR-174 and the test.

Together with #858 this takes the corpus to **zero line citations**. The two left in `Adr/Index.rst` are the convention's own examples of what not to write.

ADR-169 alone carried **137**, 92 of them bare `:NNN` continuations. Converting each one means reading what it points at, and that is where the value turned out to be.

## Three citations no longer said what the record claims

**Section 7 enumerates the 40 `denyNonAdmin()` sites**, and one entry had drifted onto the wrong method:

| | |
|---|---|
| Record said | `ToolPlaygroundController` `:155`, `:294`, `:367` |
| The three gates are at | `:156`, `:295`, `:381` |
| `:367` resolves to | `resumeAction()` — already the second entry |
| So the record | named `resumeAction()` twice, never named `submitInputAction()`, and still totalled three |

The list now names all 40 by method, and it **reconciles exactly** against the grep the section opens with:

```
real denyNonAdmin() sites:      40
named in the record:            40
  in code, missing from record: —
  in the record, not in code:   —
```

**`UseCasePackController` was cited at `:129`**, which is inside `showAction()`; `installAction()` begins at `:141`. The prose named the right method — the line number was the only wrong part, which is rather the point.

**`ext_conf_template.txt:132` is `mcpOperationTimeout`**, cited for the routing policy mode. That key is `routing.policyMode` at `:139`. A citation seven rows off landed on an unrelated setting and read as precise.

And the method names throughout were uniformly missing their `Action` suffix — `testConnection` for `testConnectionAction` — so **not one of them would have resolved** as a symbol.

## Most citations were deleted, not converted

Same finding as #858, more pronounced here:

```rst
:php:`DataHandler::isTableAllowedForThisPage()` (``DataHandler.php:7484-7503``)
```

The symbol is right there. Section 4's seven field citations are the clearest case — `` ``system_prompt_data_class`` (``:222``) `` — the field name is a better anchor than any line, and it is already in the sentence. ADR-140's six and ADR-174's one are all this shape.

## The escape hatch that emptied, and PHPStan's objection

Twenty of ADR-169's citations reach into TYPO3 core. Those become symbols too: a core symbol survives a patch release, a core line number does not, and this repository does not pin the dependency whose line numbers those were.

That left the test's declared list of unverifiable citations — sixteen entries — completely **empty**, and PHPStan refused the result at level 10:

```
Call to function in_array() with arguments mixed, array{} and true
will always evaluate to false.
```

It is right, and the fix is not a suppression. The list, the branch consulting it and the test reconciling it are gone. `everyCitationNamesAFileThatStillExists` is now absolute and says what to do instead in its own failure message. **An escape hatch nothing reaches for is a declaration nothing reads** — and the convention that replaced it is what made it unreachable.

## One thing that is load-bearing and looks incidental

The suite globs `Adr*.rst`, so `Adr/Index.rst` is not scanned. That is what lets the convention page quote `ResumeCoordinator.php:204` and `AgentRunController.php:267` as examples of rotted citations. Widening the glob to `*.rst` would fail the build on the page documenting the rule — now stated in the test docblock so nobody "fixes" it.

## One caution for anyone repeating this

I first tried to resolve all 137 mechanically, carrying the "current file" forward across the bare `:NNN` continuations. **That mapping was wrong for all 44 citations in section 7** — the numbers index the controller classes named in the surrounding prose, not the file the previous citation named, and the script happily resolved them against `Configuration/Backend/Modules.php` and produced 40 plausible `#controllerActions` anchors. The automation is usable as a lookup table; it is not usable as an applier, because the correct target is in the sentence, not in the syntax.

## Verification

| gate | result |
|---|---|
| `-s unit` (full) | 7285 tests, 24592 assertions, exit 0 |
| `-s phpstan` | No errors |
| `-s cgl -n` | SUCCESS |
| line citations in `Adr*.rst` | **0** |
| anchors / `:ref:` / repo symbols | all resolve |
| control: reintroduce `ExtensionConfiguration.php:150` | fails, naming the file and what to write instead |

### DIFF
--- a/Documentation/Adr/Adr140EffectivePolicyReadoutWithoutApplyPath.rst
+++ b/Documentation/Adr/Adr140EffectivePolicyReadoutWithoutApplyPath.rst
@@ -67,7 +67,7 @@ A read-only view, in an existing module, reading through the runtime resolvers.
    resolver exposes provenance, and it is not reconstructable from the stored
    configuration either: TYPO3's own
    :php:`ExtensionConfiguration::synchronizeExtConfTemplateWithLocalConfigurationOfAllExtensions()`
-   (``ExtensionConfiguration.php:197-218``) merges every ``ext_conf_template.txt``
+   merges every ``ext_conf_template.txt``
    default into ``settings.php`` whenever an unknown key is read or the Install
    Tool is entered. By the time anything could ask, the shipped defaults are
    already stored values.
@@ -92,21 +92,21 @@ The only API for writing extension configuration is
 :php:`ExtensionConfiguration::set()`. Three properties, each verified against
 the shipped core:
 
-- **It is ``@internal``** (``ExtensionConfiguration.php:150``) and documented as
-  *"Set a full extension configuration"* — it takes the whole array and calls
-  :php:`setLocalConfigurationValueByPath('EXTENSIONS/' . $extension, $value)`
-  (``:166``). There is no per-key write.
+- **It is ``@internal``** and documented as *"Set a full extension
+  configuration"* — it takes the whole array and calls
+  :php:`setLocalConfigurationValueByPath('EXTENSIONS/' . $extension, $value)`.
+  There is no per-key write.
 - **It materialises every shipped default.** A caller has no source for the
   array other than :php:`get()`, which returns the template-merged result. Our
-  own :php:`DataClassEnforcementDefaultUpdateWizard::executeUpdate()`
-  (``Classes/Updates/DataClassEnforcementDefaultUpdateWizard.php:77-85``) does
+  own :php:`DataClassEnforcementDefaultUpdateWizard::executeUpdate()` does
   exactly this: ``get()`` → mutate one key → ``set()``. Every default in
   ``ext_conf_template.txt`` becomes an explicitly stored value.
-  :php:`storedEnforcement()` (``:121-131``) — the "did the operator ever choose
+  :php:`storedEnforcement()` — the "did the operator ever choose
   a mode?" probe :ref:`ADR-115 <adr-115>`'s wizard depends on — could then never
   return ``null`` again.
-- **``additional.php`` spoils it.** The core docblock says so outright
-  (``ExtensionConfiguration.php:145-146``): if that file overwrites a setting,
+- **``additional.php`` spoils it.** The core docblock on
+  :php:`ExtensionConfiguration::set()` says so outright: if that file
+  overwrites a setting,
   ``->set()`` "may not end up as expected". An apply button would report success
   and the next request would serve the old value — the worst failure mode a
   governance UI can have.
--- a/Documentation/Adr/Adr169RecordManagementUsesTypo3Permissions.rst
+++ b/Documentation/Adr/Adr169RecordManagementUsesTypo3Permissions.rst
@@ -50,17 +50,17 @@ its enforcement point (its Decision, "Grants, not roles"). That rule exists
 because :ref:`ADR-023 <adr-023>` shipped backend-group checkboxes that gated
 nothing and :ref:`ADR-117 <adr-117>` had to remove them: "a control that has to
 be labelled 'has no effect' is worse than its absence"
-(``Adr117WithdrawCapabilityPermissions.rst:78-79``).
+(:ref:`ADR-117 <adr-117-decision-alternatives>`).
 
 Issue `#768` asks seven questions that have to be answered together before that
 surface is built. This record recommends answers and settles none of them.
 
 **Version scoping.** Every claim about core reach is read off the TYPO3 version
-resolved in this worktree — 14.3.5 (``Typo3Version.php:22``). The blocking
+resolved in this worktree — 14.3.5 (:php:`Typo3Version`). The blocking
 matrix is ``^13.4 || ^14.3`` (``composer.json``). 13.4 reaches the same
 capability through :php:`BackendUtility::isRootLevelRestrictionIgnored()`,
-deprecated in 14.0 in favour of the Schema API
-(``BackendUtility.php:2963-2967``); its call sites were not enumerated here.
+deprecated in 14.0 in favour of the Schema API; its call sites were not
+enumerated here.
 
 .. _adr-169-decision:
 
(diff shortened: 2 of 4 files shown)


CybotTM:

**What the gate cannot see here, and what I did instead.** A docs PR that rewrites 137 citations can silently point them somewhere else, and only one of the four resulting forms is machine-checkable. So every citation was resolved against the tree before it was written: each `:php:` symbol verified to exist as `function <name>`, each anchor with `grep -F`, each `:ref:` against the corpus's actual labels. That pass is what produced the three findings in the body — none of them came from reading the prose.

**The one I would flag to a reviewer as the load-bearing claim:** the section-7 enumeration now reconciles 1:1 against `grep -rn 'denyNonAdmin()) instanceof' Classes/Controller/` — 40 sites in the code, 40 named, nothing extra in either direction. Before this change it named `resumeAction()` twice and `submitInputAction()` never, while still totalling three. That reconciliation is a script I ran, not an impression.

**A dead end worth knowing.** I first tried to resolve all 137 mechanically. The mapping was wrong for all 44 citations in section 7 — the numbers index the controller classes named in the surrounding prose, not the file the previous citation named — and the script produced 40 plausible-looking anchors against the wrong file. Anything that reads like automation output in this diff was checked by hand afterwards.

**The guard change was controlled.** Emptying the declared list of unverifiable citations is a weakening unless something still fails: reintroducing `ExtensionConfiguration.php:150` fails with a message naming the file and what to write instead. Run, observed, restored.

**PHPStan objected to the result and was right.** With the list empty, `in_array()` against `array{}` is provably always false. The list, the branch consulting it and the test reconciling it are gone rather than suppressed — an escape hatch nothing reaches for is a declaration nothing reads.

Stacked on #858, so `queue=false` until that one merges. Not merging on this record — I hold no merge authorisation for this PR.

CybotTM:
Self-review: 83b35361a1456e34f2287e62f9a6d273e205e848

PR #862, opened 2026-08-21
security(js): make the sanitizer rule an error, and earn it

### LINKED ISSUE #854: Twelve innerHTML assignments the sanitizer rule cannot see through escapeHtml()

CybotTM:
Closes #854. #825 shipped `no-unsanitized` as a **warning** because switching it on reported 12 assignments and none of them was a hole. This makes both rules errors.

## A control failed, and it changed the plan

The issue proposed declaring `escapeHtml()` to the rule and suppressing what remained. That clears 4 of 12. I suppressed the rest, then ran the control that any relaxation deserves — **remove `escapeHtml()` from a real call site and check the gate notices**:

```
escapeHtml bei safeName entfernt:  >>> UNENTDECKT <<<
```

The rule reports **one finding per assignment**. A suppression on a template therefore hides every *other* reason that template could be unsafe — including a deliberately removed escape. That is the argument against suppressing a template, and it is why five of the six planned suppressions are gone.

After the rewrite, the same control:

| | |
|---|---|
| `escapeHtml` removed at `safeName` | **fails** |
| `escapeHtml` removed in the review list | **fails** |
| `escapeHtml` removed inside the capability accumulator | **fails** |
| markdown input no longer escaped (the one remaining suppressed site) | still undetected — stated below |

## What the rule cannot follow — measured, not assumed

Each of these was isolated with a minimal reproduction before any production code moved:

| shape | verdict | what the code does now |
|---|---|---|
| `const` declared inside a `catch` block | not traced | call the escaper at the assignment (2 error paths) |
| `.map().join('')` | opaque in every form — declaring `join` as an unwrapper does **not** help | accumulate with `+=`, which it does follow (3 sites) |
| conditional whose test is a member expression, **both branches literals** | rejected | `let` + `if` (4 in one template) |
| `Number.parseFloat` / `parseInt` output | not recognised | escaped anyway — the values are numbers, but an escaper the rule can verify beats a comment it cannot |
| `.replace()` chain, escaper called in the same expression | breaks the trace | **suppression, unavoidable** |

The ternary finding explains a dead end worth recording: **four** of them sat in one template, and because the rule reports once, removing any single one never changed the verdict. Individual bisection said "not this one" about all four. Only enumerating the interpolations and toggling them one at a time found them.

## The one remaining blind spot, and what was done about it

`renderMarkdownOutput()` keeps its suppression. Its escaping **moved from the caller into the method**, so breaking the contract now means editing the lines directly under the comment explaining it, rather than a line eighty rows away in a different method. The comment states what the suppression stops detecting, because it does stop detecting it.

Three suppressions remain for `innerHTML` round-trips — read the element's own markup, show a spinner, put it back — each with its reason.

## The rewrite is behaviour-preserving, and that was checked

Old and new template code, rendered against the same inputs — empty description, `null` capabilities, `contextLength` 0, markup in the model name, quotes in the description:

```
IDENTISCH fuer alle 3 Faelle (Whitespace normalisiert)
```

## Two things that predate this change

- **`renderOutput()`'s docblock was wrong.** It claimed `plain/markdown/json` are escaped through `escapeHtml()`. `plain` and `json` write through `textContent` and never were — correct, but not for the stated reason.
- **The sandboxed iframe is load-bearing and nothing checks it.** `renderHtmlOutput()` puts the model's raw output into `srcdoc`, which is safe *only* because `iframe.sandbox = ''` denies scripts and same-origin. `no-unsanitized` does not cover `srcdoc`, so removing that line would not fail the build. Now said at the line.

## Verification

| | |
|---|---|
| `npx eslint` | exit 0 — 0 errors, 2 pre-existing `no-unused-vars` warnings |
| `npm ci` then lint | exit 0 |
| controls | 3 of 4 removed escapes now fail the gate; the 4th is the stated blind spot |
| render equivalence | identical output, old vs new, 3 inputs incl. edge cases |

No PHP touched, so the PHP matrix is unaffected; `js-lint` is the gate that reads this change.

### DIFF
--- a/Resources/Public/JavaScript/Backend/ModelIdField.js
+++ b/Resources/Public/JavaScript/Backend/ModelIdField.js
@@ -432,8 +432,12 @@ import { readAjaxError } from '@netresearch/nr-llm/Backend/AjaxError.js';
             })
             .finally(function () {
                 button.disabled = false;
-                // Restore the original server-rendered button content (icon + text)
-                button.innerHTML = originalHTML; // eslint-disable-line no-unsanitized/property
+                // Restore the original server-rendered button content (icon +
+                // text) read from this same element before the spinner. Nothing
+                // external enters; the rule treats any value read out of
+                // innerHTML as unsafe regardless of where it came from.
+                // eslint-disable-next-line no-unsanitized/property
+                button.innerHTML = originalHTML;
             });
     }
 
--- a/Resources/Public/JavaScript/Backend/SetupWizard.js
+++ b/Resources/Public/JavaScript/Backend/SetupWizard.js
@@ -263,10 +263,13 @@ class SetupWizard {
             }
         } catch (e) {
             document.getElementById('models-loading').style.display = 'none';
-            // SECURITY: Escape error message to prevent XSS (unchanged pattern; transport-only change)
-            const discoverError = escapeHtml(await readAjaxError(e));
+            // SECURITY: escape the error before it reaches innerHTML. Called
+            // here rather than through a const: the rule cannot follow a
+            // binding declared inside a catch block, and an escaper it can see
+            // beats one it has to be told about.
             document.getElementById('models-list').innerHTML =
-                '<div class="alert alert-danger">Failed to discover models: ' + discoverError + '</div>';
+                '<div class="alert alert-danger">Failed to discover models: '
+                + escapeHtml(await readAjaxError(e)) + '</div>';
         }
     }
 
@@ -278,11 +281,15 @@ class SetupWizard {
             // SECURITY: Escape all external data to prevent XSS
             const safeName = escapeHtml(model.name);
             const safeModelId = escapeHtml(model.modelId);
-            const safeDescription = model.description ? escapeHtml(model.description) : '';
+            const safeDescription = escapeHtml(model.description || '');
 
-            const capabilities = (model.capabilities || ['chat']).map(cap =>
-                `<span class="badge bg-secondary">${escapeHtml(cap)}</span>`
-            ).join('');
+            // Accumulated rather than `.map().join('')`: the rule cannot see
+            // through map/join and would report the assignment below, which
+            // would then also hide a genuinely removed escapeHtml() there.
+            let capabilities = '';
+            for (const cap of (model.capabilities || ['chat'])) {
+                capabilities += `<span class="badge bg-secondary">${escapeHtml(cap)}</span>`;
+            }
 
             let contextStr = '-';
             if (model.contextLength) {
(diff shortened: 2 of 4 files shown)


CybotTM:
Self-review: 4f70ae63a9e0ea9524adce51523d08a5d470bb00

PR #863, opened 2026-08-21
feat(mcp): import a catalogue from the CLI, so a deploy needs no click

### LINKED ISSUE #836: nrllm:mcp:import — a CLI entry point for McpImportService, so a seeded server gets its catalogue on deploy
## Why

An MCP server record can be seeded (typo3-demo does, for `mcp.deepwiki.com`), but its catalogue can only be imported by a person pressing **Import catalogue** in the MCP Servers module. For an instance that is rebuilt by a deploy (`make update` re-applies the seed on every run), that leaves one manual step after the first deploy that carries the server — and a fresh install has no tools for that server until somebody clicks. `McpImportService::import(McpServerRecord)` already does the whole job; it just has no CLI entry point.

## What

A console command `nrllm:mcp:import <identifier>` (and `--all` for every enabled server) that runs the same `McpImportService::import()` the module runs, prints the `McpImportReport` (tools imported / skipped with reason / errors) and exits non-zero when the import failed. Respect the operation budget (ADR-170) and the SSRF gate exactly as the module does — the command is the module's button, not a second path. Admin-only is not a question on the CLI; the run is the operator's.

## Acceptance

- `vendor/bin/typo3 nrllm:mcp:import deepwiki` against a seeded row imports the catalogue and prints the report; a second run is idempotent (same count, no duplicates).
- `--all` walks every enabled server; a failing server does not stop the others, and the exit code reflects any failure.
- Unit test for the command's argument handling and exit codes with a faked import service; documented in `McpServers.rst` ("How it works", step 2) and `Documentation/Administration/Commands` if such a page exists.

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_0124WZK5UoAM9oYQoSYF1Y8z)_

CybotTM:
Closes #836. A seeded MCP server has a record and no tools until somebody opens the module and presses **Import catalogue**. On an instance rebuilt by a deploy that is a manual step after every fresh install, and the server supplies nothing until it happens.

```bash
vendor/bin/typo3 nrllm:mcp:import deepwiki
vendor/bin/typo3 nrllm:mcp:import --all
```

## It is the module's button, not a second path

Checked before writing anything: `McpServerController::importAction()` resolves the record and formats the report — that is all it does. Every refusal reason, the SSRF gate, the ADR-170 operation budget and the catalogue reconciliation live in `McpImportService::import()`. So the command holds **no policy of its own**, and the issue's "not a second path" is a property of the code rather than an intention.

## Three decisions the issue did not settle

**Identifier, not uid** — a uid is not knowable to whoever writes the deploy script. This needs no new repository method: `findEnabled()` is filtered in the command. That matters, because identifiers are deliberately **not unique** in the table — soft-deleted rows keep theirs, which is exactly why the service refuses on a twin instead of relying on an index. The ambiguity is therefore reachable here too, and is reported once naming both rather than resolved by picking a row.

**`--all` walks `findEnabled()`, not `findUsable()`** — the latter filters out servers without a data class. Silently skipping one is worse than importing it and printing the service's reason, which names the missing setting.

**An empty installation exits zero.** A deploy runs `--all` unconditionally; "no enabled server yet" is a valid state, not a red pipeline. A *refusing* server does fail the run, and does not stop the others.

## Registered explicitly, because `schedulable` is a decision

`autoconfigure: true` and the `../Classes/*` resource import would have picked up `#[AsCommand]` on its own. The entry exists so the flag is chosen rather than defaulted: this is the only one of the four registered commands that is **schedulable**, because the import is idempotent and a catalogue changes on the server's schedule rather than ours.

## Functional, and not by preference

The importer and both repositories are `final readonly`, so a faked import service is not constructible. More to the point: **the acceptance criterion that matters most — a second run writes nothing — cannot be shown by a fake that never writes a row.** Only the HTTP client is scripted, which is what the neighbouring `McpImportServiceTest` does and says why.

One trap worth naming: `McpTestServer` answers a *queue*, one entry per contact. A test importing two servers must script two, and getting that wrong reads as a refusal from the second server rather than as an exhausted fixture — which is how the `--all` test failed first. The queue depth is now an explicit argument at every call site.

## Three controls, each observed

| defect injected | result |
|---|---|
| `break` out of the loop on the first refusal | fails — the healthy server was no longer imported |
| return SUCCESS regardless of refusals | fails — exit-code assertion |
| silently take the first of two twins | fails — ambiguity assertion |

Restored tree: 10 tests green.

## An architecture rule said no, and it was right

`Classes/Command/` is "core"; the MCP importer is the tool module. `ModuleSeamTest` rejected the dependency (ADR-090), and the fix is the established one — three commands are already named in its exception list because they are a module's operational surface living in a shared directory. This is the fourth, and where it moves in a split (`nr_llm_tools`) is recorded both in the test and in the command's own docblock, as that docblock demands.

**Found while doing it:** the docblock said "**Seven** classes are excluded BY NAME" and listed **six** — a class left the list without the number following. The count is gone rather than corrected, for the reason #793 is about.

## Verification

| gate | result |
|---|---|
| `-s functional -d sqlite` (this class) | 10 tests, 52 assertions, exit 0 |
| `-s unit` (full) | 7288 tests, 24766 assertions, exit 0 |
| `-s phpstan` (carries phpat) | No errors |
| `-s cgl -n` | SUCCESS |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- **`nrllm:mcp:import` imports an MCP catalogue from the CLI**, so a seeded server gets its tools on deploy instead of waiting for somebody to press **Import catalogue** in the module. Takes a server identifier, or `--all` for every enabled server; prints imported, skipped and orphaned counts per server plus the reason for each skipped tool, and exits non-zero when a server refused. It is the module's button rather than a second path — the same refusals, the same guard rails and the same reconciliation, so a second run against an unchanged catalogue writes nothing. With `--all` a refusing server does not stop the others, and an installation with no enabled server is not a failure (#836).
+
 ### Fixed
 
 - **A merge conflict marker shipped in the ADR index.** `Documentation/Adr/Index.rst` ended on `||||||| c6503022` — a hand-resolved merge kept both sides and left the diff3 base marker behind as the file's last line, so the toctree carried a stray entry through v0.32.0. Nothing failed: the other guards read structure (status fields, citations, a rendered page) and a stray line at the end of a toctree is none of those to any of them. `NoConflictMarkersTest` now reads the bytes, and both marker shapes were watched failing before it was trusted.
--- a/Classes/Command/ImportMcpCatalogueCommand.php
+++ b/Classes/Command/ImportMcpCatalogueCommand.php
@@ -1,0 +1,198 @@
+<?php
+
+/*
+ * Copyright (c) 2025-2026 Netresearch DTT GmbH
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+declare(strict_types=1);
+
+namespace Netresearch\NrLlm\Command;
+
+use Netresearch\NrLlm\Domain\ValueObject\McpImportReport;
+use Netresearch\NrLlm\Domain\ValueObject\McpServerRecord;
+use Netresearch\NrLlm\Service\Tool\Mcp\McpImportService;
+use Netresearch\NrLlm\Service\Tool\Mcp\McpServerRepository;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * Imports an MCP server's advertised catalogue from the CLI.
+ *
+ * A server record can be seeded, but until this existed its catalogue could
+ * only arrive by a person opening the MCP Servers module and pressing **Import
+ * catalogue**. An instance rebuilt by a deploy therefore had one manual step
+ * after every fresh install, and no tools for that server until somebody
+ * clicked.
+ *
+ * This is the module's button, not a second path: it calls the same
+ * {@see McpImportService::import()} and adds nothing. Every refusal reason, the
+ * SSRF gate, the operation budget (ADR-170) and the catalogue reconciliation
+ * live in that service, which is why this class holds no policy of its own.
+ *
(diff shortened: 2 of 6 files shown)


CybotTM:

**The premise I checked before writing a line.** The issue says the command must be "the module's button, not a second path". That is only true if the controller adds nothing, so I read `McpServerController::importAction()` first: it resolves the record and formats the report, and every refusal, the SSRF gate, the ADR-170 budget and the reconciliation live in `McpImportService::import()`. The command therefore holds no policy — a property of the code, not an intention I wrote down.

**Three decisions the issue left open, each with its reason in the body:** identifier rather than uid and no new repository method; `findEnabled()` rather than `findUsable()` so a server without a data class gets the service's reason instead of silence; and exit 0 on an installation with no enabled server, because a deploy runs `--all` unconditionally.

**Functional rather than unit, and not by preference.** The importer and both repositories are `final readonly`, so a faked import service is not constructible — and the acceptance criterion that matters most, that a second run writes nothing, cannot be shown by a fake that never writes a row.

**Three controls, each observed:** `break` on the first refusal fails the test that the healthy server was still imported; returning SUCCESS regardless of refusals fails the exit-code test; silently taking the first of two twins fails the ambiguity test.

**One trap I hit and left documented in the test.** `McpTestServer` answers a *queue*, one entry per contact. A test importing two servers must script two, and getting it wrong reads as a refusal from the second server rather than as an exhausted fixture — which is how the `--all` test failed first. The queue depth is now an explicit argument at every call site.

**Rector: this is the PR where I paid for skipping my own note.** I ran cgl/phpstan/unit/functional at 8.4 and reported Rector as CI's to judge; CI failed on one `NewlineBeforeNewAssignSetRector` line. A `composerUpdate -p 8.2` afterwards reproduced it locally in a minute and confirmed the fix, and the gate is green there now.

`ModuleSeamTest` names the command by class, because `Classes/Command/` is core and the MCP importer is the tool module — the fourth such exception, with its split destination recorded as that test demands. Its docblock said "Seven classes are excluded BY NAME" while listing six; the count is gone rather than corrected.

Not merging on this record — I hold no merge authorisation for this PR.

CybotTM:
Self-review: 04ec820c31b35362bcf8b62181a97df7e33c4d00

CybotTM:
Self-review: 4774c87b1007c33efb893c70057abc9d4dfbd6c2

PR #865, opened 2026-08-21
feat(tools): reference an existing file from a content element

### LINKED ISSUE #834: attach_file_to_content_element — a purpose-built writer that references an existing FAL file from a content element
## Why

The backend chat (nr_mcp_agent) already uploads files into FAL through TYPO3's own API — `ChatApiController::fileUpload()` validates the MIME type with finfo, runs the document extractors, and stores the file via `ResourceStorage::addFile()` in a per-user upload folder — but only as a chat *attachment* (vision input, text extraction). Nothing lets the assistant make such a file part of the site: there is no writer that creates a `sys_file_reference`. The `files` group is read-only by design (ADR-047 `FalStorageGate`), and the only FAL write is `set_file_alternative_text` on an existing metadata row. Demo content with images (NEXT-153 follow-up, and the `ai_filemetadata` showcase that starts from an image without alt text) therefore stops at the upload.

## What

A seventh purpose-built writer, on the terms of the six existing ones (ADR-135/146/180): `attach_file_to_content_element`.

- **Input:** `content_element` (uid, required), `file` (uid of an existing `sys_file`, required), `field` (one of an allow-list the tool intersects with the live TCA of the element's CType — `assets`, `image`, `media`; default chosen from the CType), optional `title`, `alternative`, `description` for the reference, optional `position` (append by default).
- **Effect:** creates ONE `sys_file_reference` row through the DataHandler as the acting user (inline relation on `tt_content.<field>`), in the live workspace only. No upload, no file move, no metadata write — the file must already exist in a storage the `FalStorageGate` allows and inside the user's file mounts; the element must be editable by the user (`CONTENT_EDIT` on its page). A file the user may not reach and an element the user may not edit are refused with the same neutral words.
- **Preview (ADR-136):** element (page, header), field, file (name, identifier), the reference's title/alternative, position.
- **Effect class:** `NON_IDEMPOTENT_WRITE` (a second call attaches the file twice).
- **Read-back:** the reference exists with `tablenames`/`fieldname`/`uid_foreign` as planned; otherwise the reference is deleted again.
- Disabled by default, `editing` group, approval on every call, editor action declared on `tt_content`.

## Not in scope

Uploading (the chat already does that through the TYPO3 API; a tool that uploads is a different risk class — see hn/typo3-mcp-server PR #102 for how large that gets), moving files, writing `sys_file_metadata` beyond what `set_file_alternative_text` does, attaching to `pages.media` (a second table, a second record type — a later tool if wanted).

## Acceptance

- Unit tests for the schema and every refusal; functional tests against the real DataHandler: reference created with the planned `foreign_match_fields`, hidden state irrelevant (a reference has none), editor without `CONTENT_EDIT` refused, file outside the allowed storages refused, read-back failure deletes the reference again.
- `BuiltinToolGroupsTest`, `EditorActionDeclarationTest`, `EditorActionIconRegistrationTest`, `ToolEffectCoverageTest`, `ProductFactsConsistencyTest` and every counting surface updated (48 / seven).
- `Tools.rst` section, README sentence, CHANGELOG.

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_0124WZK5UoAM9oYQoSYF1Y8z)_

CybotTM:
Closes #834. The seventh purpose-built writer, and the first that creates a `sys_file_reference`. Demo content with images stopped at the upload: nothing could make an uploaded file part of the site, because the `files` group is read-only by design (ADR-047) and the only FAL write was `set_file_alternative_text` on a metadata row that already existed.

## A control caught my own design, and the design lost

The tool wrote through the DataHandler **twice**, and I had a measurement to justify it: a probe showed a single-pass write leaving the element's counter one short and the new reference sorting *first*. Both issue comments say so.

Then I disabled the second pass and **every test still passed**.

The probe had run without `$GLOBALS['LANG']`. In that environment the defect is real; in a complete backend environment — the only one this tool will run in, because `refuseWithoutBackendEnvironment()` checks — a single datamap carrying the parent list and the `NEW_` placeholder produces the right count *and* the right `sorting_foreign`. One pass now, `substNEWwithIDs` only to identify the row for the read-back, and the docblock says what was measured rather than what I remembered.

The issue comments are corrected rather than left standing.

## Three refusals the description did not name

Each came from the live TCA, not from reasoning about it:

| | |
|---|---|
| **The field must accept the file.** | `image` takes fourteen extensions, `assets` twenty-seven, `media` anything. A `.docx` on `image` is a relation the FormEngine would reject, so it is refused rather than written. |
| **`sys_language_uid` must be in the payload.** | Without it a non-admin's write dies inside `checkRecordEditAccess()` on a missing `languageField`. Stated rather than defaulted, which also pins the tool to the default language like the other FAL writer. |
| **`title`, `alternative`, `description` are `exclude` fields.** | The DataHandler drops them for a user without the grant — silently, empty `errorLog`. Without checking them the tool reports a caption it did not write. |

That last one is also what made the read-back **testable**: remove the grant and the guard fires.

## Discarding is a rollback, not a delete

When the read-back fails, removing the reference row is not enough — the element's counter still names it, which is exactly the inconsistent state this tool exists to avoid producing. `discard()` now writes the parent field back to the list that was there before the call. A test caught that too.

## Six controls, each observed

| defect injected | result |
|---|---|
| read-back stops checking the texts | fails |
| `discard()` stops restoring the counter | fails |
| extension check removed | fails |
| ambiguity branch removed | fails |
| storage gate removed | fails |
| page-permission check removed | fails |

Restored tree: 13 tests green. An earlier run of this same harness reported two of these as *undetected*; both were failed `sed` anchors, so the test had run against an unmodified file. The harness now compares the checksum before and after and says "anchor missed" instead of "undetected" — a control that cannot fail is worse than none.

## One measured fact worth recording

**No CType that ships with the core offers two of `image` / `assets` / `media`.** `textpic` and `image` show `image`, `textmedia` shows `assets`, nothing shows two. The "several fields → name the one you mean" branch is therefore unreachable with core content types, so the test registers a two-field type of its own rather than leaving the branch an untested claim.

There is no `position` argument: the write appends by construction, and an argument nothing reads is worse than none.

## Verification

| gate | result |
|---|---|
| `-s functional -d sqlite` (this class) | 13 tests, 86 assertions, exit 0 |
| `-s functional` (tool surfaces) | 32 tests, 193 assertions, exit 0 |
| `-s unit` (full) | 7289 tests, 24795 assertions, exit 0 |
| `-s phpstan` | No errors |
| `-s cgl -n` | SUCCESS |
| `-s rector -n` (pinned 8.2, run locally) | Rector is done |

Every counting surface moves with it: **47 tools → 48**, **six writers → seven**, across `Tools.rst`, `README.md`, seven landing-page data files in both languages, and the four tests that pin those numbers.


CybotTM:
Review record — no Copilot review exists (monthly, account-wide quota), so the green gate is not a read of this change. This is a writing tool touching FAL and the DataHandler, so the record is longer than usual.

**The finding I would most want a reviewer to check, because it overturns my own design.** The tool wrote through the DataHandler twice, justified by a probe showing a single pass leaving the counter short and the reference sorted first. Disabling the second pass changed **no test**. The probe had run without `$GLOBALS['LANG']` — an environment the tool refuses to run in. One pass now. If anyone disagrees, the disagreement is with a measurement that is reproducible in either direction: add the language service and the defect disappears.

**Authorisation, which is the part worth being paranoid about.** Two independent gates against the *acting* user, both verified by controls that remove them:

- The element's page needs `CONTENT_EDIT` (`doesUserHaveAccess`).
- The file goes through `FalStorageGate::isFileAccessible()`. I read that method rather than trusting its name: it applies the extension's storage allow-list, then builds the storage through `createFromRecord()` with `setEvaluatePermissions(true)`, the user's `filePermissionsFor()` and each of their mounts, then asserts with `checkFileActionPermission('read', …)`. One call is the whole check, mounts included.

Both refuse in the same words as "does not exist", so a refusal never confirms a uid — asserted by a test that runs an absent file and an unreachable one and compares the strings.

**The silent-drop class, which is where the sibling writer was bitten.** `title`, `alternative` and `description` are `exclude` fields on `sys_file_reference`. Without the field grant the DataHandler drops them and logs nothing, so the tool would have reported a caption it never wrote. The read-back checks them; removing the grant makes it fire, and the test asserts the reference is removed again afterwards.

**Discarding is a rollback.** Deleting the row alone leaves the element's counter naming a reference that is gone — the exact inconsistency this tool exists to avoid. A test caught it.

**Six controls, each observed** (read-back texts, counter restore, extension check, ambiguity branch, storage gate, page permission). Worth naming: an earlier run of this harness reported two as *undetected*, and both were failed `sed` anchors — the test had run against an unmodified file. The harness now compares checksums and says "anchor missed". **A control that cannot fail is worse than none**, and I nearly reported two blind spots that did not exist.

**Unreachable branch, made reachable in the test rather than left as a claim.** No core CType offers two of `image`/`assets`/`media` — measured. The ambiguity refusal therefore cannot fire on core content, so the test registers its own two-field type.

**Counting sweep**: 47 → 48 tools and six → seven writers, across `Tools.rst`, `README.md`, seven landing-page data files in both languages and four tests. Verified by re-querying every surface for a stale number afterwards, not by trusting the replace count. The ADR files that also match `\b47\b` were checked and left alone — those are line citations and ADR-047 references, not tool counts.

Gates run locally at 8.2 including Rector, plus the full unit suite. Not merging on this record — I hold no merge authorisation for this PR.

CybotTM:
Self-review: 6c7d198886f55f8ac464641624f2277c4d64a078

CybotTM:
Self-review: 5dbcab129fcf7beea815ceaf8f94a908aca13c2a

CybotTM:
Self-review: f0b47adf9fc656f11435c9c7c4626b20eb3db18e

PR #866, opened 2026-08-21
fix(telemetry): attribute what the specialized services spend

### LINKED ISSUE #844: Specialized services (image, speech, transcription, DeepL) cannot carry caller source — withCallerSource() compiles there and is dropped

CybotTM:
Closes #844.

The specialized services — image generation, speech synthesis, transcription, DeepL — spend money the per-extension breakdown could not attribute, while their options objects *looked* annotatable. A consumer could call `withCallerSource()` on them, the call compiled, and the value was silently dropped. That is worse than an absent feature: nr-landingpage left its AI image generation deliberately unannotated for exactly this reason (netresearch/nr-landingpage#42).

**One correction to the issue before the fix.** #844 gives two reasons and the second is wrong. It says the services "call `runLifecycle()` directly rather than going through the manager's middleware pipeline", so a preserved value would never reach the middleware. They do go through the pipeline — `runLifecycle()` *is* `$this->pipeline->run(...)`, `TelemetryMiddleware` reads `$context->metadata['sourceExtension']`, and `UsageMiddleware` already aggregates the cost row on that same key. What is missing is not a reader. It is that these services build their own `ProviderCallContext` instead of going through `LlmServiceManager`, which is the only place that translates the identity into that metadata. So the readers were waiting and were never asked. That made the fix considerably smaller than the issue implies: no middleware changed.

**What changed**

- `AbstractOptions::withCallerSourceFromArray()` — re-applies an identity that arrived in a consumer-supplied array. The four `fromArray()` implementations rebuild through the constructor and lost anything the base class held.
- `AbstractSpecializedService::callerSourceMetadata()` for the services with a typed options DTO, and `::callerSourceFromArray()` for those taking a plain array (FAL and DeepL have no DTO). Both produce the keys the pipeline already reads. The array variant lives on the base class rather than in each service because `extractBeUserUid()` is already duplicated three times in that namespace and a fourth copy of the same idea is not an improvement — the existing duplication is left alone rather than fixed in a bug fix.
- Wired into DALL·E, FAL, Whisper, TTS and DeepL at the point where each builds its dispatch context.

**The asymmetry is deliberate and now asserted.** `fromArray()` reads the consumer's own input and carries the identity; `toArray()` builds the provider payload and must not — the caller's extension key is attribution metadata, exactly like `configuration` and the budget fields already absent there for the same reason. `toArrayNeverExposesTheCallerIdentity` asserts that absence per option class, so a later change that makes the two "symmetric" for tidiness cannot quietly start sending the calling extension's name to OpenAI, FAL or DeepL.

**The guards were watched failing, not just passing.** With the four `fromArray()` changes reverted, 8 of the 20 round-trip cases fail — precisely the ones asserting the identity survives — while the leak guards stay green, which is correct since they describe behaviour that must hold before and after. With only the TTS wiring reverted, exactly one test fails: `anAnnotatedCallReachesTheMiddlewareWithItsCallerIdentity`. Both were restored and re-run green.

**Deliberately not covered**

- `DallEImageService::variation()` and `::edit()` take no options object. They stay unattributed rather than gaining a parameter with no consumer — the public signatures are in the frozen API surface and nothing asked for this.
- DeepL's language detection and its usage/languages metadata calls are internal sub-steps, not consumer-attributable spend.

**Gates run locally** (PHP 8.4 × one matrix cell — CI covers eight): `cgl` clean, `phpstan` level 10 clean, full `unit` suite 7310 tests green, API-surface and changelog checks green. The functional suite was not run locally; nothing here touches persistence.

No ADR: this implements ADR-177 and ADR-178 on a path they already claimed rather than deciding anything new. The one decision worth recording — that `toArray()` stays asymmetric — is stated in the code and pinned by a test.


CybotTM:
Self-review: 5f55038ed6c953c38b3408f35a9ec333228e05e8

PR #869, opened 2026-08-21
fix(docs): a merge conflict marker shipped in the ADR index

CybotTM:
`Documentation/Adr/Index.rst` ends on this, on `main`, and shipped that way in **v0.32.0**:

```
   Adr180TheSixthWriterCreatesAPage
   Adr181EventStreamFramedAnswers
||||||| c6503022
```

[`41bb809a`](https://github.com/netresearch/t3x-nr-llm/commit/41bb809a) — "Merge branch 'main' into fix/mcp-accept-sse" — resolved a conflict by hand, kept both sides, and dropped only the `<<<<<<<` and `>>>>>>>` lines. The diff3 **base** marker stayed as the file's last line, so the ADR toctree carries a stray entry. `c6503022` is [#829](https://github.com/netresearch/t3x-nr-llm/pull/829), the other side of that merge.

One occurrence repo-wide; found by accident while reading the index for something else.

## It is published, not just committed

`https://docs.typo3.org/p/netresearch/nr-llm/0.32/en-us/Adr/Index.html` answers **200** and contains, verbatim:

```html
<p>||||||| c6503022</p>
```

Sphinx did not read the stray line as a broken toctree entry and drop it — it rendered it as an ordinary **paragraph after the list**. That is why the `docs` job was green: there was no unresolved reference to complain about. The page builds without error and shows a conflict marker to anyone reading the ADR index for the released version.

So "it renders and returns 200" was not evidence here either. Same shape as the guards one level down: the check reads structure, the defect is bytes.

## Why nothing caught it, which is the argument for the guard

Every existing check on this corpus reads **structure**:

| check | reads |
|---|---|
| `AdrLifecycleTest` | status vocabulary, `:Amends:`/`:Amended:` pairing |
| `AdrReferenceIntegrityTest` | that `:ref:` targets resolve |
| `AdrCodeCitationTest` | citations against the tree |
| the `docs` job | a rendered page |

A stray line at the end of a toctree is none of those things to any of them. The suites passed, the docs job rendered, the release went out. **Only a check that reads the bytes finds it**, so that is what `NoConflictMarkersTest` does — repo-wide, not just the ADR corpus.

## Two controls, each observed

| injected | reported |
|---|---|
| the shipped marker put back | `Documentation/Adr/Index.rst:536  \|\|\|\|\|\|\| c6503022` |
| `<<<<<<< HEAD` appended to `README.md` | `README.md:375  <<<<<<< HEAD` |

Clean tree: 3 tests green.

## `=======` is deliberately not a marker, and there is a test saying so

Seven equals signs at the start of a line are how RST underlines a heading. "Completing" the marker list would turn every heading in the corpus into a finding, so the omission is pinned by an assertion rather than left as a comment somebody helpfully fixes later.

The base marker is the one that matters anyway: a resolution that forgets it leaves a file that **no longer looks conflicted at a glance**, which is precisely how this one got through review and a release.

## Verification

| gate | result |
|---|---|
| `-s unit` (full) | 7291 tests, 24770 assertions, exit 0 |
| `-s phpstan` | No errors |
| `-s cgl -n` | SUCCESS |
| `-s rector -n` (pinned 8.2) | Rector is done |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- **A merge conflict marker shipped in the ADR index.** `Documentation/Adr/Index.rst` ended on `||||||| c6503022` — a hand-resolved merge kept both sides and left the diff3 base marker behind as the file's last line, so the toctree carried a stray entry through v0.32.0. Nothing failed: the other guards read structure (status fields, citations, a rendered page) and a stray line at the end of a toctree is none of those to any of them. `NoConflictMarkersTest` now reads the bytes, and both marker shapes were watched failing before it was trusted.
+
 ## [0.32.0] - 2026-08-21
 
 ### Fixed
--- a/Documentation/Adr/Index.rst
+++ b/Documentation/Adr/Index.rst
@@ -533,4 +533,3 @@ Tools
    Adr179ADroppedForcedSourceIsRecordedOnTheRun
    Adr180TheSixthWriterCreatesAPage
    Adr181EventStreamFramedAnswers
-||||||| c6503022
--- a/Tests/Unit/NoConflictMarkersTest.php
+++ b/Tests/Unit/NoConflictMarkersTest.php
@@ -1,0 +1,138 @@
+<?php
+
+/*
+ * Copyright (c) 2025-2026 Netresearch DTT GmbH
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+declare(strict_types=1);
+
+namespace Netresearch\NrLlm\Tests\Unit;
+
+use PHPUnit\Framework\Attributes\CoversNothing;
+use PHPUnit\Framework\Attributes\Test;
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+use SplFileInfo;
+
+/**
+ * No file in this repository carries a merge conflict marker.
+ *
+ * `Documentation/Adr/Index.rst` shipped one in v0.32.0. A hand-resolved merge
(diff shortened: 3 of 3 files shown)


CybotTM:

**The severity is verified, not relayed.** A peer session reported the marker was live on docs.typo3.org; I checked it myself rather than repeating it: `curl` answers **HTTP 200**, 82414 bytes, exactly one occurrence of `<p>||||||| c6503022</p>`. That is what turns this from repo hygiene into a published defect, so it belongs in the body with the evidence.

**Both controls were watched failing**, in the position each exists for: putting the shipped marker back reports `Documentation/Adr/Index.rst:536`; a `<<<<<<< HEAD` line appended to `README.md` reports that instead. Clean tree green afterwards, and the second control was re-run once more after the amend.

**The `=======` omission is pinned by an assertion, not a comment.** Seven equals signs are how RST underlines a heading. A later contributor "completing" the marker list would turn every heading in the corpus into a finding, so the omission is a test.

**One thing I got wrong and fixed, stated because it is the same class the PR is about.** CI's Code Style job failed on the first push. My first instinct was a fixer-version difference between the Docker run and the runner — I checked instead of assuming: both are `friendsofphp/php-cs-fixer v3.95.20`, confirmed against `installed.json` and the CI log. The real cause was mine: a hook reformatted the test file **after** my gate run, and I committed the hook's version without re-running the gate. Order is fix → `-n` → commit again, and every gate was re-run on the final tree rather than on the one I had measured.

That is worth saying here because it is the same shape as the defect: a check that ran, then something changed underneath it, and the result was reported as if it still held.

Scope: one line removed, one test added, one CHANGELOG entry. No PHP behaviour touched.

Not merging on this record — I hold no merge authorisation for this PR. Once it is on `main` the published 0.32 docs still carry the marker until a 0.32.1 is cut; that is a release decision, not part of this change.

CybotTM:
Self-review: 76ff6e43bde8803fc76ebbf7541d9b55bb0f4a6d

PR #870, opened 2026-08-21
fix(agent): keep the caller source across resume and the queue

### LINKED ISSUE #847: A resumed or queued agent run loses its caller source
The caller source survives only as long as the `AgentRunRequest` is in memory. Everything that rehydrates a run from persisted state gets it back as null, so the continuation's provider calls are attributed to nobody.

The chain: `AbstractOptions` keeps `callerSourceExtension` / `callerSourceOperation` deliberately out of `toArray()` (documented at `AbstractOptions.php:101-105` — it is call metadata, not a provider option). `AgentRunRequestCodec::encode()` persists the options as `$request->options?->toArray()`, so the pair is not written. `AgentRunRequestCodec::decode()` rebuilds them with `ToolOptions::fromArray()`, which does not read them either, and the codec re-attaches only `plannedCost` and `idempotencyKey` afterwards. `ToolLoopService::resumeAfterApproval()` and the input-submission twin do the same thing from the suspended state (`ToolOptions::fromArray($state->options, $beUserUid)`, lines 622 and 713).

Consequences for a consumer that annotates its calls correctly:

- `AgentRuntimeInterface::approve()` takes no options object at all, and the persisted state carries none either, so every provider round-trip after an approval is unattributed. For an agent whose writing tools all suspend for approval (ADR-134), that is the majority of the interesting traffic.
- `submitInput()` has the same shape.
- A queued run executed by `runQueued()` goes through the codec, so it is unattributed for its whole length even though the enqueuing request named its caller.

The effect is worse than a missing row: the same logical run appears partly under its extension and partly under *Unattributed*, so neither number is the run's cost.

The two obvious fix shapes, both nr-llm-internal: persist the pair explicitly in the codec and the suspended state next to `plannedCost` / `idempotencyKey` (keeping `toArray()` provider-clean), or carry them on the run row itself since the telemetry row already has the columns.

Found while annotating netresearch/t3x-nr-mcp-agent#134 (PR netresearch/t3x-nr-mcp-agent#135), which can name its chat turn but not its approval continuation; verified by reading nr-llm v0.31.1 as installed.

CybotTM:
Closes #847. Sibling of #866; the two are independent and can land in either order.

The caller identity survived only as long as the `AgentRunRequest` was in memory. `AgentRunRequestCodec::dehydrate()` persists the options as `$request->options?->toArray()`, and `toArray()` omits the caller source by design; `rehydrate()` rebuilds them with `ToolOptions::fromArray()`, which does not read it either. `ToolLoopService::resume()` and its input-submission twin do the same from the suspended state. So a run that named its caller lost it the moment it was persisted.

The effect is worse than a missing row. The same logical run shows up partly under its extension and partly under *Unattributed*, so **neither number is the run's cost**. For an agent whose writing tools all suspend for approval (ADR-134), the unattributed part is most of the interesting traffic.

**Where it is carried.** Beside the serialised options, not inside them — the channel the codec already uses for `plannedCost` and the idempotency key, with the rationale already written down there. Putting it inside `toArray()` would have sent the calling extension's key upstream as a provider parameter, and it would have taken the reason those other two are excluded down with it. That argument is the peer session's, not mine, and it is better than the one I had.

**What changed**

- `AgentRunRequestCodec` — the pair is written next to `plannedCost` / `idempotencyKey` and re-injected on rehydrate in the same shape.
- `ToolLoopService::persistedRunOptions()` — what goes into the suspended state at both suspend sites (approval and input pauses).
- `ToolLoopService::restoreCallerSource()` — used by both resume paths, mirroring how they already re-inject the acting user's uid by hand (ADR-084).

`SuspendedRunState` is untouched: it is frozen `@api`, and the pair rides in the `options` array it already carries. `ToolOptions::fromArray()` ignores keys it does not know, so the two survive the round trip without reaching a provider, and resume reads them back explicitly.

`AgentRuntimeInterface::approve()` still takes no options object — it does not need one. The identity now comes from the persisted state, which is where an approval's context has always come from.

**The guards were watched failing.** With `AgentRunRequestCodec` reverted, exactly one test fails: `roundTripCarriesTheCallerIdentityOutOfBand`. With `ToolLoopService` reverted, exactly one fails: `anApprovedRunContinuesUnderTheCallerThatStartedIt`. In both cases the paired absence tests stay green, which is correct — they assert behaviour that must hold before and after. Both were restored and re-run green.

The end-to-end assertion is the one that matters: a run is started with `withCallerSource('nr_mcp_agent', 'chatTurn')`, suspends for approval, and the options the manager receives on the *resumed* call are checked for the identity — not the helper in isolation.

**Gates run locally** (PHP 8.4 × one matrix cell — CI covers eight): `cgl` clean, `phpstan` level 10 clean, full `unit` suite 7292 tests green. The functional suite was not run locally; the persisted shape is exercised by the codec's round-trip test rather than against a database.

No ADR: this restores ADR-177/ADR-178 on a path they already claimed. The decision worth recording — that the identity travels beside `toArray()` and not inside it — is stated at both writers and pinned by tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - **A merge conflict marker shipped in the ADR index.** `Documentation/Adr/Index.rst` ended on `||||||| c6503022` — a hand-resolved merge kept both sides and left the diff3 base marker behind as the file's last line, so the toctree carried a stray entry through v0.32.0. Nothing failed: the other guards read structure (status fields, citations, a rendered page) and a stray line at the end of a toctree is none of those to any of them. `NoConflictMarkersTest` now reads the bytes, and both marker shapes were watched failing before it was trusted.
+- **A resumed or queued agent run keeps the caller that started it.** The identity survived only while the `AgentRunRequest` was in memory: `AgentRunRequestCodec` persisted the options through `toArray()`, which omits the caller source by design, and the resume paths rebuilt them the same way. Every provider round-trip after an approval or an input submission was therefore unattributed, and a queued run was unattributed for its whole length even though the enqueuing request named its caller. That is worse than a missing row — the same logical run appeared partly under its extension and partly under *Unattributed*, so neither figure was the run's cost. For an agent whose writing tools all suspend for approval (ADR-134) it was most of the interesting traffic (#847).
+
+  The pair now travels beside the serialised options rather than inside them, in the channel the codec already uses for `plannedCost` and the idempotency key, and the resume paths re-inject it exactly as they re-inject the acting user's uid (ADR-084). `toArray()` stays provider-clean: widening it would have sent the calling extension's key upstream and taken the reason those other fields are excluded down with it.
 
 ## [0.32.0] - 2026-08-21
 
--- a/Classes/Service/Agent/AgentRunRequestCodec.php
+++ b/Classes/Service/Agent/AgentRunRequestCodec.php
@@ -76,6 +76,14 @@ final readonly class AgentRunRequestCodec
             // request hit the identical budget gate and dedup.
             'plannedCost'      => $request->options?->getPlannedCost(),
             'idempotencyKey'   => $request->options?->getIdempotencyKey(),
+            // Same channel, same reason (#847): the caller identity is call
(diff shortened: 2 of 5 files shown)


CybotTM:
Self-review: 985ff6f841940fc34d819d7dd3ca465b360ee0eb

PR #871, opened 2026-08-21
docs(adr): propose the write-target contract #772 is blocked on

CybotTM:
Refs #772. **`Proposed`, not `Accepted`** — this names a decision to take, and the implementation lands after it rather than inside it, as `Classes/AGENTS.md` requires for a contract change.

## The gap, stated by the code rather than by me

ADR-176 shipped the explicit half of the per-call outcome and found the observed half blocked on a fact nothing persists: **which record a tool write produced**. `CallOutcome::isImplemented()` returns false for `ACCEPTED_UNCHANGED`, `EDITED` and `DISCARDED` — a checkable statement, not a sentence in a record. #772 is that gap.

ADR-122 is where it was deferred, and its reason was good:

> A contract designed before the first writing tool guesses at the shape that tool needs, and this codebase has just spent three changes removing exactly that kind of guess.

**Seven writers exist now** (`grep -rl 'use WritesThroughDataHandlerTrait' Classes/Service/Tool/Builtin/`), and the outcome signal is the first reader the contract lacked. The shape can be read off what those seven already do.

## Scope: identity, and nothing else

A table and a uid. Carried on `ToolResult`, persisted through `RunTrace` against the run's existing `correlation_id`, read by the observed outcome joining `sys_history`.

**The other two things ADR-122 deferred stay deferred, with their reasons intact** — that is deliberate, and it is the part most likely to be "helpfully" widened later:

| deferred | still deferred because |
|---|---|
| idempotency scope | `ToolEffect` already tells the reaper what may be retried; no caller asks a finer question |
| preview contract | ADR-136 shipped `ToolPreviewInterface` for the approval card, driven by arguments |

## The place this will be lost, named before it is

`ToolLoopService` does not return the tool's result. It **rebuilds** it:

```php
return $result->isError
    ? ToolResult::error($this->bounder->content($result->content))
    : ToolResult::text(
        $this->bounder->content($result->content),
        ...$this->bounder->artifacts($result->artifacts),
    );
```

A new property on `ToolResult` would be dropped on every path, **and every test that asserts on the tool's own return value would still pass.** The tool would declare a write target, the loop would answer with one that has none, and the outcome signal would find nothing.

This is the **third** time this shape has cost something here — [#845](https://github.com/netresearch/t3x-nr-llm/issues/845) and [#846](https://github.com/netresearch/t3x-nr-llm/issues/846) lost values to an options rebuild; [#844](https://github.com/netresearch/t3x-nr-llm/issues/844) found `withCallerSource()` callable on the specialized services and dropped by `fromArray()`. In each case the field existed, the reader existed, and the rebuild in between answered without it.

So the acceptance asks for a test that runs a writer **through the loop**, not one that calls `execute()` directly — which is the assertion that would have passed in all three earlier cases.

## Consequences worth reading before accepting

- `ToolResult` is on the frozen surface. The change is **additive**, and the surface file has to be regenerated in the same change — a consequence to accept, not an obstacle to work around.
- Seven tools gain one line each, and a writer that forgets it reports no outcome silently. `ToolEffectCoverageTest` already knows which tools declare a write effect, so the same list is what makes the omission fail.
- The join reads `sys_history`, which has its own retention (ADR-064). An outcome derived after the history row was purged is **unknown**, not "no change", and has to be recorded as neither.

ADR-122 carries the counterpart `:Amended:` field; `AdrLifecycleTest` enforces the pairing in both directions.

## Verification

| gate | result |
|---|---|
| `-s unit` (full, incl. `AdrLifecycleTest` and `AdrReferenceIntegrityTest`) | 7288 tests, 24782 assertions, exit 0 |
| `-s phpstan` | No errors |
| `-s cgl -n` | SUCCESS |
| line length | every line ≤ 80 |

## Correction: this is stacked on #869, and the reason is worth reading

I first wrote here that this PR and [#869](https://github.com/netresearch/t3x-nr-llm/pull/869) "merge cleanly in either order", because one adds a toctree line and the other removes the line below it. **I then tested that claim instead of leaving it, and it was false.** The two lines are adjacent, git cannot decide, and the merge conflicts:

```
CONFLICT (content): Merge conflict in Documentation/Adr/Index.rst
```

The resolution left behind — before anyone touches it — is this:

```
   Adr182AWriteToolNamesTheRecordItWrote
||||||| c6503022
=======
>>>>>>> origin/fix/adr-index-conflict-marker
```

**Three markers, in the file whose single stray marker is the entire subject of #869.** A careless resolution here reproduces the defect being fixed, three times over, in the same place.

So this branch is now rebased onto `fix/adr-index-conflict-marker` and targets it. When #869 merges, GitHub retargets this to `main`. The guard from #869 runs in this branch and passes — 0 markers, one `Adr182` entry.

### DIFF
--- a/Documentation/Adr/Adr122ToolEffectContractDeferred.rst
+++ b/Documentation/Adr/Adr122ToolEffectContractDeferred.rst
@@ -8,7 +8,7 @@ ADR-122: The side-effecting tool contract waits for a side-effecting tool
 
 :Status: Accepted (premise expired — see :ref:`ADR-135 <adr-135>` and :ref:`ADR-136 <adr-136>`)
 :Date: 2026-07-29
-:Amended: 2026-08-09 by :ref:`ADR-135 <adr-135>` and :ref:`ADR-136 <adr-136>`
+:Amended: 2026-08-09 by :ref:`ADR-135 <adr-135>` and :ref:`ADR-136 <adr-136>`; 2026-08-21 by :ref:`ADR-182 <adr-182>`
 :Authors: Netresearch DTT GmbH
 
 .. note::
--- a/Documentation/Adr/Adr182AWriteToolNamesTheRecordItWrote.rst
+++ b/Documentation/Adr/Adr182AWriteToolNamesTheRecordItWrote.rst
@@ -1,0 +1,159 @@
+.. include:: /Includes.rst.txt
+
+.. _adr-182:
+
+============================================================================
+ADR-182: A write tool names the record it wrote
+============================================================================
+
+:Status: Proposed
+:Date: 2026-08-21
+:Amends: :ref:`ADR-122 <adr-122>` (the deferral whose premise expired)
+:Authors: Netresearch DTT GmbH
+
+.. _adr-182-context:
+
+Context
+=======
+
+:ref:`ADR-176 <adr-176>` made a per-call outcome recordable and shipped the
+explicit half: a person rates a result, the row is written against the call's
+``correlation_id``. It also found the observed half blocked on something it did
+not expect to look for — **nothing persists which record a tool write
+produced**. The uid the model chose sits inside
+:php:`ToolInvocation::$arguments` as free-form JSON, and
+:php:`ToolEffectInterface` declares only :php:`getEffect()`, a classification
+rather than an identity. Without a
+queryable write target there is no join to ``sys_history``, so
+``ACCEPTED_UNCHANGED``, ``EDITED`` and ``DISCARDED`` cannot be derived.
+
+:php:`CallOutcome::isImplemented()` says so in code rather than in prose, and
+returns false for exactly those three. `#772` is that gap.
+
+:ref:`ADR-122 <adr-122>` is where the missing fact was deferred. It declined to
+build a side-effecting tool contract with a plain reason: "A contract designed
+before the first writing tool guesses at the shape that tool needs, and this
+codebase has just spent three changes removing exactly that kind of guess." Its
+``:Status:`` already records that premise as expired.
+
+**Seven writers exist now** — ``grep -rl 'use WritesThroughDataHandlerTrait'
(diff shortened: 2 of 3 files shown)


CybotTM:

**What a gate cannot check here at all.** This is a `Proposed` ADR: the tests confirm the lifecycle fields pair up and the `:ref:` targets resolve, and nothing more. Whether the *decision* is right is the thing a reader has to supply. So the record was written against the code rather than from memory, and the three load-bearing claims are each traceable:

- **Seven writers** — `grep -rl 'use WritesThroughDataHandlerTrait' Classes/Service/Tool/Builtin/`. Not a remembered number; the same anchor corrected ADR-176's "five" earlier today.
- **`ToolResult` is on the frozen surface** — line 927 of `Tests/Unit/Api/api-surface.txt`. `ToolEffectInterface`, `ToolPreviewInterface` and `EditorActionInterface` are not, which is why the record proposes carrying the fact on the result rather than on a new interface method.
- **The loop rebuilds the result** — `ToolLoopService:1135-1140`, quoted verbatim. That is the section of the record I would most want challenged, because it is the part that turns an obvious design into a trap.

**The verdict above is against the post-rebase head**, checked rather than assumed: 75 success / 18 skipped / 0 failure on `27f1613e`, which is what `gh pr view` reports as the head. A green run from before a force-push describes a tree that no longer exists.

**One thing in this PR was wrong and is corrected in the body rather than quietly fixed.** I claimed this branch and #869 "merge cleanly in either order". I then ran the merge: it conflicts, and the resolution git leaves behind carries **three** conflict markers in the very file whose single marker is #869's whole subject. The branch is now stacked on #869 and the claim is replaced by the measurement. That check happened only because I had written the claim down publicly — which is the argument for writing them down.

Not merging on this record — I hold no merge authorisation for this PR, and it should not land before #869 in any case.

CybotTM:
Self-review: 27f1613eb58f15c7ba70127b410f40030f9bfabb

PR #872, opened 2026-08-21
chore(spec-kit): keep the specifications, drop the generated scaffolding

CybotTM:
`#824` vendored two Spec Kit presets into `.specify/`. Reviewing what that actually put under version control turned up three problems, and the obvious fix for them does not exist. This keeps the part of Spec Kit this project wrote and drops the part the CLI generates.

## What was wrong with the committed state

**It is generated.** `speckit.manifest.json` carries a SHA for every script and template, because the CLI writes them and can rewrite them. A committed copy drifts from its source the moment spec-kit releases.

**It pins one coding agent.** `.specify/init-options.json` records `"ai": "claude"` — in a repository whose agent documentation is otherwise neutral.

**It claims a state nobody has.** `.specify/presets/.registry` and `.specify/integrations/claude.manifest.json` are versioned and between them list thirteen installed commands and skills under `.claude/`. That directory is in `.gitignore`, is empty in all three worktrees here, and cannot be anything else in a fresh clone. Both files were committed assertions that were already false.

## Why not declare the presets as a dependency instead

That was the first idea, and it was measured rather than assumed:

| Question | Answer |
|---|---|
| Can a preset be pinned to a tag or branch? | Yes — a catalog entry's `download_url` points at `archive/refs/tags/v1.zip` or `refs/heads/main.zip` |
| Is the origin recorded on install? | **No.** `specify_cli/presets/__init__.py:3583` hardcodes `"source": "local"` whether the preset came from a local directory, a tagged URL or a catalog |
| Is there a restore-from-declaration command? | **No.** `init, check, version, self, extension, integration, event, preset, bundle, workflow` — none restores |
| Can CI verify it? | **No.** `specify preset info <id>` reads the *catalog*, exiting 0 for a preset that is merely listed and not installed |

`.specify/preset-catalogs.yml` declares only where presets may be *found*, never which are installed — a repository entry with no `require` block. And the state that would matter lives in `.claude/`, which is per-workstation and gitignored, so nothing in this repository can observe it.

## What stays

`specs/` and `.specify/memory/constitution.md` — written here, readable without any tool, and the constitution path is already cited by `specs/001-per-call-cost-and-tokens/plan.md`. The ignore rule blocks the directory and exempts the constitution; both directions verified with `git check-ignore`.

`AGENTS.md` gains the setup line for anyone who wants the generated commands locally, and states plainly that nothing verifies they ran it. `#824` is the evidence that this works: its specification was written against the templates by hand, with no command installed anywhere.

The `.specify` skip entries in `check-agent-doc-symlinks.php` and `AgentDocsCountConventionTest` stay — the directory still exists locally for anyone who installs the presets, and both checks would otherwise sweep the vendored `AGENTS.md` files again.

## Verification

`make gate` passed end to end: cgl, PHPStan, unit, fuzzy, Rector pinned to 8.2, functional against SQLite, plus the CHANGELOG check. All four `ci:test:repo` scripts pass.

One pre-existing warning is unrelated to this change: `verify-harness.sh` reports a build/CI drift heuristic against the last commit, and reports it identically on `main`.

## Not in this change

The two preset repositories have no tags at all — `v0.1.0` exists only inside their `preset.yml`. Until they cut releases, neither a catalog nor a documented `--from` URL can point at a stable artifact. That is their repositories' work, not this one's.


CybotTM:
Copilot was unavailable — the review quota is exhausted for the month and account-wide, so re-requesting cannot change it before the reset. Reviewed by hand instead.

**What the diff removes, and how I checked it is safe to remove.** 52 deletions. 43 of them are covered directly: 32 are the vendored preset copies, and 11 appear by SHA in `speckit.manifest.json`, which exists precisely because the CLI writes those files and can rewrite them. The remaining seven — `init-options.json`, `integration.json`, both integration manifests, `.constitution-template.json`, and the two workflow files — are in no manifest, so I did not take them on trust. A throwaway `specify init --here` in an empty directory produced exactly that set. They are init output, not authored content.

The one file in `.specify/` this project wrote, `memory/constitution.md`, stays, and the ignore rule was verified in both directions with `git check-ignore`: the generated paths are ignored, the constitution is not. Its path is load-bearing — `specs/001-per-call-cost-and-tokens/plan.md` cites it.

**Why a dependency declaration is not the alternative.** That was the first idea and it does not exist. `specify_cli/presets/__init__.py:3583` hardcodes `"source": "local"` into the registry whatever the install path was, so the origin is never recorded; no command restores an installation from a declaration; and `specify preset info` reads the catalog rather than the installed set — it exits 0 for a preset that is merely listed, which I confirmed against `a11y-governance` in a project where it was not installed. My first attempt at that check ran through `head` and reported the exit code of `head`, which nearly produced the opposite conclusion.

**What changed since the PR was opened.** Both preset repositories now have a `v0.1.0` release, so the setup line in `AGENTS.md` points at `archive/refs/tags/v0.1.0.zip` instead of `--dev <local path>`. Exercised: a fresh `specify init` project installed both from those URLs and `specify preset list` reports each as `v0.1.0 — enabled`. Before tagging I checked that `git archive` of each tree carries `preset.yml` and `commands/` and carries neither symlink, since `safe_extract_archive` rejects any archive containing one — a tag is a poor place to discover that.

**Verification.** `make gate` passed end to end twice, most recently on `3d1b9725`: cgl, PHPStan, unit, fuzzy, Rector pinned to 8.2, functional against SQLite, plus the CHANGELOG check. All four `ci:test:repo` scripts pass, and `verify-harness.sh` moves from `Level 2 PARTIAL` to `Level 3 COMPLETE` here — the drift warning it reports on `main` is about the last commit touching CI files without `AGENTS.md`, which this branch does touch.

**What this review does not establish.** I did not test that a contributor following the `AGENTS.md` line ends up with working slash commands in an agent other than Claude Code; the install was exercised with `--integration claude` only. And nothing in this repository can verify that anyone ran the setup at all — that is the point of the change rather than an oversight, but it means the workflow's first step rests on the rule being followed, not on a check.

CybotTM:
Self-review: 3d1b9725c5df5626d73396d510f19f1964c0033c

